PR #4, opened 2021-10-09
add resnet50 from paddlehub conversion

fengyuentau:
- [x] ResNet and demo implementation
- [x] Benchmark support and results for ResNet


fengyuentau:
cc @kaingwade @zihaomu

fengyuentau:
Need to update `download_data.py` with test images for image classification.

Issue #10, opened 2021-11-02
face_detection_yunet failed to load pretrained model as default


atinfinity:
##### System information (version)
- OpenCV => 4.5.4
  -  <https://pypi.org/project/opencv-python/4.5.4.58/>
- opencv_zoo => <https://github.com/opencv/opencv_zoo/commit/6a66e0d6e47a693e6d0dd01bbb18e920f3fbae75>
- Operating System / Platform => Ubuntu 20.04 64bit
- Compiler => gcc 9.3.0
- Python => 3.8.10

##### Detailed description

The explanation of [models/face_detection_yunet/README.md](https://github.com/opencv/opencv_zoo/blob/6a66e0d6e47a693e6d0dd01bbb18e920f3fbae75/models/face_detection_yunet/README.md) is as follows.

```shell
# detect on camera input
python demo.py
# detect on an image
python demo.py --input /path/to/image
```

The `face_detection_yunet/demo.py` try to load `face_detection_yunet.onnx` as default.  
But, this pretrained model was renamed to `face_detection_yunet_2021sep.onnx` by <https://github.com/opencv/opencv_zoo/pull/7>.

As a result, `face_detection_yunet/demo.py` failed to load pretrained model as default.

```shell
$ python3 demo.py
Traceback (most recent call last):
  File "demo.py", line 60, in <module>
    model = YuNet(modelPath=args.model,
  File "/home/opencv/opencv_zoo/models/face_detection_yunet/yunet.py", line 22, in __init__
    self._model = cv.FaceDetectorYN.create(
cv2.error: OpenCV(4.5.4-dev) /tmp/pip-req-build-tpkxoqhj/opencv/modules/dnn/src/onnx/onnx_importer.cpp:198: error: (-5:Bad argument) Can't read ONNX file: face_detection_yunet.onnx in function 'ONNXImporter'
```

##### Steps to reproduce

```shell
$ git lfs clone https://github.com/opencv/opencv_zoo.git
$ cd opencv_zoo/models/face_detection_yunet
$ python3 demo.py
```

fengyuentau:
Thanks for the feedback. Models are renamed https://github.com/opencv/opencv_zoo/pull/7, but I forgot to change demo scripts accordingly.

fengyuentau:
https://github.com/opencv/opencv_zoo/pull/11 is created to correct default values of models filenames in demo scripts. You can help review changes.

PR #11, opened 2021-11-02
Correct default values of model filenames in demo scripts

fengyuentau:
related issue https://github.com/opencv/opencv_zoo/issues/10


fengyuentau:
cc @kaingwade @zihaomu

atinfinity:
@fengyuentau Thank you for your mention. <https://github.com/opencv/opencv_zoo/issues/10#issuecomment-957132129>
I think that there is a concern. Because, the filename of pretrained model has date(e.g. face_detection_yunet_`2021sep`.onnx).
So, you will need to modify demo scripts when you update model.

fengyuentau:
@atinfinity I think that's not a big problem since models won't be updated quite frequently.

PR #27, opened 2021-11-23
Improve hardware namings on the table of benchmark results

fengyuentau:
Basically rename hardware setups to `<board/brand_name>-<processing_unit>`, such as `RPI-CPU` for the ARM CPU on Rasberry Pi, `INTEL-CPU` for the Intel CPU.


fengyuentau:
cc @kaingwade @zihaomu

PR #28, opened 2021-11-24
Add hardware setup: TB-RV1126Ds CPU

fengyuentau:
cc @kaingwade @zihaomu


fengyuentau:
CPU spec: https://developer.arm.com/ip-products/processors/cortex-a/cortex-a7

fengyuentau:
We do not have plans to add the benchmark results on this CPU into the table for now.

Issue #31, opened 2021-12-12
face_detection_yunet


donaldlee2008:
os:windows
python: 3.6
opecv-python:4.5.4
face_detection_yunet/demo.py   
cmd: python demo.py   -i 3.jpg
error:
onnx_importer.cpp:203: error: (-210:Unsupported format or combination of formats) Failed to parse ONNX model: face_detection_yunet_2021sep.onnx in function 'cv::dnn::dnn4_v20211004::ONNXImporter::ONNXImporter'

fengyuentau:
Did you test on WIndows CMD or WSL?

donaldlee2008:
> Did you test on WIndows CMD or WSL?

WIndows CMD

donaldlee2008:
It works when I download the model from 
https://github.com/ShiqiYu/libfacedetection.train/blob/a61a428929148171b488f024b5d6774f93cdbc13/tasks/task1/onnx/yunet.onnx   
but face_detection_yunet_2021sep.onnx does not

bwmcin:
Opened face_detection_yunet_2022mar.onnx in notepad+ and got this...

version https://git-lfs.github.com/spec/v1
oid sha256:50ef07f702a31741ca46a4c0d947773b64143b9362780237bf0d427d6c79bab7
size 345478

....and nothing else.  Also, the file size is very small.

face_detection_yunet_2022_mar-act_int8-wt_int8-quantized.onnx has similar issue.

fengyuentau:
@bwmcin You need to install git-lfs from https://git-lfs.github.com/ and run 
```shell
cd your/opencv_zoo
git lfs install
git pull
```
Or clone a fresh copy of opencv_zoo after you install git-lfs:
```shell
git lfs install
git clone https://github.com/opencv/opencv_zoo
```

zihaomu:
Hi @donaldlee2008, does `git lfs install` solve your problem?

deepti-pushpak:
cv2.error: OpenCV(4.6.0) /io/opencv/modules/dnn/src/onnx/onnx_importer.cpp:260: error: (-210:Unsupported format or combination of formats) Failed to parse ONNX model: face_detection_yunet_2022mar.onnx in function 'ONNXImporter'

I am having the same problem on my ubuntu 20.04.4. Could anyone resolve this?

fengyuentau:
[...]

PR #45, opened 2022-03-31
Use YuNet of fixed input shape to avoid 'parseShape' error

fengyuentau:


zihaomu (approved):
LGTM :+1:

bwmcin:
I ran the latest version of face_detection_yunet on AGX, with OpenCV 4.5.5, with backend CUDA and target CUDA, and received error -220, unsupported format or combination of formats, onnx_importer.cpp:199.  I was using the face_detection_yunet_2022mar.onnx file.  Don't know if this is related to above or not.

fengyuentau:
@bwmcin There is a tracking issue https://github.com/opencv/opencv_zoo/issues/31 for the simular problem. Let me reproduce it and see what the problem is.

PR #49, opened 2022-04-20
Update README.md

zihaomu:
Add `per-tensor` or `per-channel` tag for the performance test of Khadas VIM3.


fengyuentau:
Very good 👍

PR #64, opened 2022-06-27
addDBfrontpage

the-star-sea:
- [x] add DB demo to frontpage


zihaomu (commented):
Thanks for your contribution. It would nice if you could provide Chinese and English gif Demo.

fengyuentau (changes requested):
I suggest you replace the example of detecting Chinese characters with another one without QR code or anything privacy sensitive.

Besides, you should also add these two images to the readme in `models/text_detection_db`. It can even better if you present the command of how you run this demo and get the result image.

the-star-sea:
> I suggest you replace the example of detecting Chinese characters with another one without QR code or anything privacy sensitive.
> 
> Besides, you should also add these two images to the readme in `models/text_detection_db`. It can even better if you present the command of how you run this demo and get the result image.

 I see there is a shell in the readme of DBnet.

fengyuentau (approved):
Looks good to me👍

zihaomu (approved):
Thank you! LGTM 👍

PR #69, opened 2022-07-06
Add scripts to evaluate models in the zoo on different datasets

fengyuentau:
- [x] Evaluation framework
- [x] Initial dataset for evaluation: ImageNet
- [x] Report mobilenet accuracy
- [x] Report ppresnet accuracy

Note: the accuracy of PP-ResNet quantized is very low, which seems to be bad quantization.

### DIFF
--- a/tools/eval/eval.py
+++ b/tools/eval/eval.py
@@ -1,0 +1,89 @@
+import os
+import sys
+import argparse
+
+import numpy as np
+import cv2 as cv
+
+from datasets import DATASETS
+
+if "PYTHONPATH" in os.environ:
+    root_dir = os.environ["PYTHONPATH"]
+else:
+    root_dir = os.path.join("..", "..")
+sys.path.append(root_dir)
+from models import MODELS
+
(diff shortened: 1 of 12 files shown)


tools/eval/eval.py
```diff
@@ -0,0 +1,89 @@
...
+
+datasets = dict(
+        imagenet=dict(
+            name="ImageNet",
+            topic="image_classification",
+            size=224),
+)
+
+def main(args):
+    # Instantiate model
```
zihaomu, line 64: At the beginning of the script we can print the model name and test data home name.

zihaomu (approved):
Thank you! 👍

tools/eval/eval.py
fengyuentau: Such information is printed in `dataset.eval()` like the following:

image

fengyuentau:
The issue of low accuracy of quantized PP-ResNet will be fixed in another pull request. Merge as is.

PR #72, opened 2022-07-13
Add script to evaluate face recognition by LFW

WanliZhong:
The script is modified based on [evaluation of InsightFace](https://github.com/deepinsight/insightface/blob/f92bf1e48470fdd567e003f196f8ff70461f7a20/src/eval/lfw.py).

- [x] Add dataset for evaluation: [LFW](http://vis-www.cs.umass.edu/lfw/)
- [x] Report sface accuracy


kaingwade:
LGTM.

tools/eval/datasets/lfw.py
```diff
@@ -0,0 +1,240 @@
...
+        img = cv.imread(self.lfw_image_paths[key])
+        if self.target_size != self.LFW_IMAGE_SIZE:
+            img = cv.resize(img, (self.target_size, self.target_size))
+        return img
+
+    def eval(self, model):
+        ids = self.ids
+        embeddings = np.zeros(shape=(len(self), 128))
+        for idx, img in tqdm(enumerate(self), desc="Evaluating {} with {} val set".format(model.name, self.name)):
+            embedding = model.infer(img)
```
fengyuentau, line 228: The image should be aligned before being fed into the model. This can be the problem of low accuracy.

To get aligned image, you need to get the bounding box of the face using YuNet. For simplicity, you can start with treating the bounding box of highest score as the only face in the image. So I suggest you:
1. Validate whether alignment using YuNet improves the accuracy.
2. If it does, store those boxes in a constant in this file (or a separate .npy file) and load it when requires to save time.

WanliZhong: Thanks! After using YuNet to get bbox, the accuracy is 97.92%. this is still not 99.6% as described in the documentation. I will try to solve this problem.

fengyuentau:
The difference can be YuNet failing to detect a face in some images. You can have a check on the images with low score faces or no face.
________________________________
发件人: Wanli ***@***.***>
发送时间: Monday, July 25, 2022 4:02:17 PM
收件人: opencv/opencv_zoo ***@***.***>
抄送: Yuantao Feng ***@***.***>; Assign ***@***.***>
主题: Re: [opencv/opencv_zoo] Add script to evaluate face recognition by LFW (PR #72)


@WanliZhong commented on this pull request.

________________________________

In tools/eval/datasets/lfw.py<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fopencv%2Fopencv_zoo%2Fpull%2F72%23discussion_r928579674&data=05%7C01%7C%7Ce9066fff51754f3eb8dd08da6e13fe99%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637943329392846011%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=aJ%2BNFzFzfqqsxhw0Jlf9XmeR3GJ4TawZj%2BnfJQsf66U%3D&reserved=0>:

> +                        self.image_path_pattern.format(person_name=person2_name, image_name=image2_name)
+                    ]
+                    id_list.append(False)
+        return image_paths, id_list
+
+    def __getitem__(self, key):
+        img = cv.imread(self.lfw_image_paths[key])
+        if self.target_size != self.LFW_IMAGE_SIZE:
+            img = cv.resize(img, (self.target_size, self.target_size))
+        return img
+
+    def eval(self, model):
+        ids = self.ids
+        embeddings = np.zeros(shape=(len(self), 128))
+        for idx, img in tqdm(enumerate(self), desc="Evaluating {} with {} val set".format(model.name, self.name)):
+            embedding = model.infer(img)


Thanks! After using YuNet to get bbox, the accuracy is 97.92%. this is still not 99.6% as described in the documentation. I will try to solve this problem.

―

tools/eval/datasets/lfw.py
fengyuentau: The difference can be YuNet failing to detect a face in some images. You can have a check on the images with low score faces or no face.

WanliZhong: It may not be the problem. Each image is detected with a face and the selected bbox scores are all above 0.8.

fengyuentau: Sometimes it may fail even with score 0.8. Please, take a look at those faces with score lower than 0.9.

fengyuentau (approved):
`lfw_face_features.npy` should be renamed to `lfw_face_bboxes.npy` to address confusions. Should also explain what this file is for and the data structure (e.g. `[[x1, y1, x2, y2, lm1_x, lm1_y, ... ]]`) in the section of LFW in readme.

Others look good to me 👍 Thank you so much!

zihaomu (approved):
LGTM! 👍

PR #71, opened 2022-07-13
Text Recognition: Add script to evaluate text recognition by ICDAR2003

Charles-258:

PR #79, opened 2022-08-02
crnnFP16

Charles-258:
add two crnn quantized models


models/text_recognition_crnn/README.md
```diff
@@ -2,11 +2,23 @@
...
 
+Results of accuracy evaluation with [tools/eval](../../tools/eval).
+
+| Models | Accuracy(%) |
+| ------ | -------------- |
+| CRNN_EN | 81.66 |
+| CRNN_EN_FP16 | 82.01 |
+| CRNN_CH | 71.28 |
+| CRNN_CH_FP16 | 78.63 |
+\*: 'FP16' stands for 'model quantized into FP16'.
```
zihaomu, line 13: We should carry experiment on more OCR datasets. Then, I think we can merge it. Please add info about the test dataset.

zihaomu: @Charles-258, friendly reminder

zihaomu (approved):
LGTM! 👍

fengyuentau:
We need to add notes in the readme telling users how to use these fp16 models and compare the accuracy with fp32 models' as well.

zihaomu:
@fengyuentau This info can be found in the merged PR #71.

fengyuentau:
Thank you but we still need notes in the readme about how to use these fp16 models with opencv.

zihaomu:
In my opinion, we will have more fp16 models in the future, we need to add a unified note about how to load float16 in OpenCV. How about adding this info in https://github.com/opencv/opencv_zoo/tree/master/benchmark/README.md? 
The main idea is to tell the user that loading float 16 ONNX model requires opencv with this patch [PR in opencv](https://github.com/opencv/opencv/pull/22337).
The best solution for python is `pip install opencv-python-rolling`.

PR #85, opened 2022-08-19
Let mp_palmdet support multiple palms detection

WanliZhong:
This PR enables `mp_palmdet` model to support multiple palm detection.
- added a `topK` parameter
- modified the return value format of the model 
```shell
from
score, bbox_coords, landmarks_coords

to
[
  [bbox_coords, landmarks_coords, score]
  ...
  [bbox_coords, landmarks_coords, score]
]
```


models/palm_detection_mediapipe/mp_palmdet.py
```diff
@@ -2,10 +2,11 @@
 import cv2 as cv
 
 class MPPalmDet:
-    def __init__(self, modelPath, nmsThreshold=0.3, scoreThreshold=0.5, backendId=0, targetId=0):
+    def __init__(self, modelPath, nmsThreshold=0.3, scoreThreshold=0.5, topK=5000, backendId=0, targetId=0):
```
fengyuentau, line 5: Need to change benchmark config when API changes.

models/palm_detection_mediapipe/mp_palmdet.py
```diff
@@ -68,17 +69,26 @@ def _postprocess(self, output_blob, original_shape):
         xy2 = (cxy_delta + wh_delta / 2 + self.anchors[:, :2]) * original_shape
         boxes = np.concatenate([xy1, xy2], axis=1)
         # NMS
-        keep_idx = cv.dnn.NMSBoxes(boxes, score, self.score_threshold, self.nms_threshold, top_k=1)
+        keep_idx = cv.dnn.NMSBoxes(boxes, score, self.score_threshold, self.nms_threshold, top_k=self.topK)
         if len(keep_idx) == 0:
-            return None, None, None
-        selected_score = score[keep_idx][0]
-        selected_box = boxes[keep_idx][0]
+            return None
```
fengyuentau, line 74: Returning an empty array would be better.

models/palm_detection_mediapipe/mp_palmdet.py
```diff
@@ -68,17 +69,26 @@ def _postprocess(self, output_blob, original_shape):
...
+            landmark += selected_anchors[idx]
+        selected_landmarks *= original_shape
 
-        return (selected_score, selected_box, selected_landmarks)
+        # [
+        #   [bbox_coords, landmarks_coords, score]
+        #   ...
+        #   [bbox_coords, landmarks_coords, score]
+        # ]
+        return np.c_[selected_box.reshape(-1, 4), selected_landmarks.reshape(-1, 14), selected_score.reshape(-1, 1)]
```
fengyuentau, line 91: Need to rerun benchmark on CPU to see whether the speed changes a lot.

models/palm_detection_mediapipe/demo.py
```diff
@@ -106,12 +111,18 @@ def visualize(image, score, palm_box, palm_landmarks, fps=None):
...
             # Inference
             tm.start()
-            score, palm_box, palm_landmarks = model.infer(frame)
+            results = model.infer(frame)
             tm.stop()
 
             # Draw results on the input image
-            if score is not None and palm_box is not None and palm_landmarks is not None:
-                frame = visualize(frame, score, palm_box, palm_landmarks, fps=tm.getFPS())
+            if results is not None:
```
fengyuentau, line 118: if `results` is an empty array, we do not need this if condition and the for loop stops.

models/palm_detection_mediapipe/demo.py
```diff
@@ -69,19 +69,24 @@ def visualize(image, score, palm_box, palm_landmarks, fps=None):
         image = cv.imread(args.input)
 
         # Inference
-        score, palm_box, palm_landmarks = model.infer(image)
-        if score is None or palm_box is None or palm_landmarks is None:
+        results = model.infer(image)
+        if results is None:
```
fengyuentau, line 73: if `results` is an empty array, we do not need this if condition and the for loop stops.

models/palm_detection_mediapipe/mp_palmdet.py
WanliZhong: After testing, there isn‘t much difference in speed on cpu before and after modification. But I found that the speed on RPI-CPU should be about 170ms instead of 89.41ms as written in the README. The time on intel-cpu is about 25ms instead of 15.5. Am I not configured correctly somewhere?

models/palm_detection_mediapipe/mp_palmdet.py
WanliZhong: After testing, there isn‘t much difference in speed on cpu before and after modification. But I found that the speed on RPI-CPU should be about 170ms instead of 89.41ms as written in the README. The time on intel-cpu is about 25ms instead of 15.5. Am I not configured correctly somewhere?

WanliZhong: Sorry I seem to have mistaken. The results of the test are not very different, do I need to update the benchmark results?

fengyuentau: no need if minor difference

models/palm_detection_mediapipe/demo.py
```diff
@@ -69,19 +69,24 @@ def visualize(image, score, palm_box, palm_landmarks, fps=None):
         image = cv.imread(args.input)
 
         # Inference
-        score, palm_box, palm_landmarks = model.infer(image)
-        if score is None or palm_box is None or palm_landmarks is None:
+        results = model.infer(image)
+        if len(results) == 0:
             print('Hand not detected')
         else:
```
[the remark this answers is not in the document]
WanliZhong: ok

models/palm_detection_mediapipe/demo.py
```diff
@@ -69,30 +69,34 @@ def visualize(image, score, palm_box, palm_landmarks, fps=None):
         image = cv.imread(args.input)
 
         # Inference
-        score, palm_box, palm_landmarks = model.infer(image)
-        if score is None or palm_box is None or palm_landmarks is None:
+        results = model.infer(image)
+        if len(results) == 0:
             print('Hand not detected')
-        else:
+        for idx, palm in enumerate(results):
```
fengyuentau, line 75: I propose to put this loop in `visualize` to keep the code clear and simple.

models/palm_detection_mediapipe/demo.py
```diff
@@ -69,30 +69,34 @@ def visualize(image, score, palm_box, palm_landmarks, fps=None):
...
+            print('-----------palm {}-----------'.format(idx))
             print('score: {:.2f}'.format(score))
             print('palm box: {}'.format(palm_box))
             print('palm_landmarks: ')
-            for plm in enumerate(palm_landmarks):
+            for plm in palm_landmarks:
                 print('\t{}'.format(plm))
 
             # Draw results on the input image
             image = visualize(image, score, palm_box, palm_landmarks)
```
fengyuentau, line 88: Put the loop in `visualize` so as to avoid calling this function in a loop

models/palm_detection_mediapipe/demo.py
```diff
@@ -106,11 +110,16 @@ def visualize(image, score, palm_box, palm_landmarks, fps=None):
...
 
             # Draw results on the input image
-            if score is not None and palm_box is not None and palm_landmarks is not None:
+            for _, palm in enumerate(results):
+                score = palm[-1]
+                palm_box = palm[0:4]
+                palm_landmarks = palm[4:-1].reshape(7, 2)
+
+                # Draw results on the input image
                 frame = visualize(frame, score, palm_box, palm_landmarks, fps=tm.getFPS())
```
fengyuentau, line 123: same as above.

fengyuentau (approved):
Other looks good to me!

PR #88, opened 2022-08-29
Use QuantFormat.QOperator by default to avoid fake quantization

fengyuentau:
Since onnxruntime==1.11.0, quantize_static.quant_format is set to QuantFormat.QDQ by default, which produces fake quantizated models. Explicitly setting it to QuantFormat.QOperator solves this problem.


zihaomu (approved):
LGTM!

WanliZhong:
LGTM!

WanliZhong (approved):

PR #89, opened 2022-08-31
text_recognition example argparse bug fix

swenkel:
The text recognition CRNN demo.py argument parser is incomplete.

Running

```
python3 demo.py --input ./examples/demo.jpg
```

yields the following error:

```
Traceback (most recent call last):
  File "opencv_zoo/models/text_recognition_crnn/demo.py", line 75, in <module>
    image = cv.resize(image, [args.width, args.height])
AttributeError: 'Namespace' object has no attribute 'width'
```

Error is caused by incomplete copy&paste of the argument parser of the `demo.py` in the text detection example.
This pull request fixes that.

### DIFF
--- a/models/text_recognition_crnn/demo.py
+++ b/models/text_recognition_crnn/demo.py
@@ -64,7 +64,7 @@ if __name__ == '__main__':
     recognizer = CRNN(modelPath=args.model, charsetPath=args.charset)
     # Instantiate DB for text detection
     detector = DB(modelPath='../text_detection_db/text_detection_DB_IC15_resnet18_2021sep.onnx',
-                  inputSize=[736, 736],
+                  inputSize=[args.width, args.height],
                   binaryThreshold=0.3,
                   polygonThreshold=0.5,
                   maxCandidates=200,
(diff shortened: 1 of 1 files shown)


zihaomu:
Hi @Charles-258, please take a look.

zihaomu (approved):
Thanks for your contribution! 👍

fengyuentau (commented):
I propose to to change line 79:
```.diff
- image = cv.resize(image, [args.width, args.height])
+ image = cv.resize(image, [736, 736])
```
instead of adding arguments, since we do not need variable width & height in the script.

swenkel:
An alternative would be changing line 63 from fixed size to using width and input from the arguments. Then it would be on par with the original source of the detector (https://github.com/opencv/opencv_zoo/blob/master/models/text_detection_db/demo.py#L65). That way the functionality and intention of the size arguments would be preserved.

fengyuentau:
> An alternative would be changing line 63 from fixed size to using width and input from the arguments.

Okay, please either make it fixed or on par with the source of detector.

fengyuentau:
Kindly note that the fixed size is set in line 113 as well.

fengyuentau (approved):
LGTM👍

PR #92, opened 2022-09-09
Add DB float16

the-star-sea:
add two DB quantized models


fengyuentau:
@the-star-sea @zihaomu What is the status of this pull request? Do you plan to continue it?

the-star-sea:
This has been done. Actually it depends on the need of your group.

the-star-sea:
Previously, zihao told me that we didn't need the function so that I transferred it as a draft.

the-star-sea:
Maybe we can simply have a chat to communicate the need and function design and I am available these days @zihaomu @fengyuentau.

zihaomu:
Hi @the-star-sea, can your change this PR to ready? 
I propose to merge this PR first. For now, pp-OCR has not been supported, this PR wants to add float16 models. It's a supplement for us.

fengyuentau:
If we are merging this PR, we need to remove the old fp32 models and replace with the fp16 models in this PR. Also need to add a note in README specifying the required minimum version of opencv that supports fp16 models.

fengyuentau:
Closing for no activities.

PR #94, opened 2022-09-22
revise the quantize-ort to quantize crnn into int8 with high accuracy

Charles-258:


tools/quantize/quantize-ort.py
```diff
@@ -29,9 +29,10 @@ def get_next(self):
 
     def get_calibration_data(self, image_dir):
         blobs = []
+        supported = ["jpg", "JPG", "png", "PNG"]  # 支持的文件后缀类型
```
zihaomu, line 32: Please change the comment to English.

zihaomu (approved):
Thanks for your contribution! 👍 

Please take look the following comment. And don't forget to squash the commits into 1 commit

Charles-258:
您好，我是王仡垚，您的邮件我已经收到，我会尽快回复。

tools/quantize/quantize-ort.py
```diff
@@ -29,9 +29,10 @@ def get_next(self):
 
     def get_calibration_data(self, image_dir):
         blobs = []
+        supported = ["jpg", "JPG", "png", "PNG"]  # supported file suffix
```
fengyuentau, line 32: use `.lower()` and keep only one instance for each format.

tools/quantize/quantize-ort.py
```diff
@@ -55,13 +56,13 @@ def __init__(self, model_path, calibration_image_dir, transforms=Compose(), per_
 
     def check_opset(self, convert=True):
         model = onnx.load(self.model_path)
-        if model.opset_import[0].version != 11:
-            print('\tmodel opset version: {}. Converting to opset 11'.format(model.opset_import[0].version))
-            # convert opset version to 11
-            model_opset11 = version_converter.convert_version(model, 11)
+        if model.opset_import[0].version != 13:
```
fengyuentau, line 59: which operator in quantized crnn needs opset 13?

zihaomu: @Charles-258 , please take a look.

Charles-258: > which operator in quantized crnn needs opset 13?

If quant_format=QuantFormat.QDQ and per_channel=True, we need to convert onnx into opset 13.
5MZ31C}UEF4`S SP%4H54W4

In other conditions, opset 11 is enough.

zihaomu: Looks like the difference is the `Quantizelinear` and `DequantizeLinear` which are very important in the QDQ mode. Quantizelinear of opset 13 has added a new attribute, [more details](https://github.com/onnx/onnx/blob/main/docs/Operators.md#quantizelinear).

fengyuentau:
It is a bad move merging with `if model.opset_import[0].version != 13:`, which means all of our quantized models need to be re-generated. It should be special-cased for the crnn models.

PR #96, opened 2022-09-27
Add resize back code in text detection and recognization demo.

zihaomu:


models/text_detection_db/demo.py
```diff
@@ -86,6 +88,7 @@ def visualize(image, results, box_color=(0, 255, 0), text_color=(0, 0, 255), isC
 
         # Draw results on the input image
         image = visualize(image, results)
+        image = cv.resize(image, [original_w, original_h])
```
fengyuentau, line 91: I propose to generate a copy for the resized image so that we do not have to resize the image back to its original size.

zihaomu: That means we need to scale the output bounding box, which is more complicated than this.

fengyuentau (approved):
👍

PR #97, opened 2022-09-28
Update benchermark and quantize

WanliZhong:
- Add the command for setting `PYTHONPATH` in powershell
- Let benchmark use the last path of environment
- Intel neural-compressor need  `onnxruntime-extensions`

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -20,12 +20,20 @@ Time is measured from data preprocess (resize is excluded), to a forward pass of
 Run the following command to benchmark on a given config:
 
 ```shell
-PYTHONPATH=.. python benchmark.py --cfg ./config/face_detection_yunet.yaml
+export PYTHONPATH=$PYTHONPATH:.. 
+python benchmark.py --cfg ./config/face_detection_yunet.yaml
 ```
 
 If you are a Windows user and wants to run in CMD/PowerShell, use this command instead:
+- CMD
 ```shell
-set PYTHONPATH=..
+set PYTHONPATH=%PYTHONPATH%;..
+python benchmark.py --cfg ./config/face_detection_yunet.yaml
+```
+
+- PowerShell
+```shell
+$env:PYTHONPATH=$env:PYTHONPATH+";.."
 python benchmark.py --cfg ./config/face_detection_yunet.yaml
 ```
 <!--
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -30,7 +30,7 @@ def prepend_pythonpath(cfg):
             prepend_pythonpath(v)
         else:
             if 'path' in k.lower():
-                cfg[k] = os.path.join(os.environ['PYTHONPATH'], v)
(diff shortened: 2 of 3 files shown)


tools/quantize/requirements.txt
```diff
@@ -1,4 +1,5 @@
 opencv-python>=4.5.4.58
 onnx
 onnxruntime
+onnxruntime-extensions
```
fengyuentau, line 4: Why do we need this package?

benchmark/benchmark.py
```diff
@@ -30,7 +30,7 @@ def prepend_pythonpath(cfg):
             prepend_pythonpath(v)
         else:
             if 'path' in k.lower():
-                cfg[k] = os.path.join(os.environ['PYTHONPATH'], v)
+                cfg[k] = os.path.join(os.environ['PYTHONPATH'].split(os.pathsep)[-1], v)
```
fengyuentau, line 33: If it is set to load the last one, I suggest we change all of the lines setting environment in REDME to
```bash
export PYTHONPATH=$PYTHONPATH:.. # for bash
#and for windows as well
```

WanliZhong: I forgot to add this setting

tools/quantize/requirements.txt
WanliZhong: OK

WanliZhong: Because there seems to be some problems with the old version, perhaps they have been solved now. I will check this again

WanliZhong: I have met the same problem as https://github.com/intel/neural-compressor/issues/94 when I only install onnxruntime. But after I install the `onnxruntime-extensions` mentioned by the maintainers from the link, the problem can be solved and jump to the next error message. Just let me check it again. I think it's my fault.

WanliZhong: I have checked this problem. The `neural-compressor` did need the `onnxruntime-extensions`, but it will not install automatically just like `onnxruntime`.

image

fengyuentau: OK, we can keep this change. Please update for the other comment, and we can merge this pr.

benchmark/README.md
```diff
@@ -20,12 +20,19 @@ Time is measured from data preprocess (resize is excluded), to a forward pass of
 Run the following command to benchmark on a given config:
 
 ```shell
-PYTHONPATH=.. python benchmark.py --cfg ./config/face_detection_yunet.yaml
+PYTHONPATH=$PYTHONPATH:.. python benchmark.py --cfg ./config/face_detection_yunet.yaml
```
fengyuentau, line 23: make it two lines:
```
export PYTHONPATH=$PYTHONPATH:..
python benchmark.py --cfg ./config/face_detection_yunet.yaml
```

WanliZhong:
All changes have been done

fengyuentau:
Please squash commits into one.

PR #100, opened 2022-10-08
Add facial expression recognition

crywang:


models/facial_expression_recognition/demo.py
```diff
@@ -0,0 +1,163 @@
...
+            pass
+
+    _boxes = copy.deepcopy(dets[:, :4])
+    keepIdx = cv.dnn.NMSBoxes(
+        bboxes=_boxes.tolist(),
+        scores=dets[:, -1].tolist(),
+        score_threshold=args.conf_threshold,
+        nms_threshold=args.nms_threshold,
+        top_k=args.top_k,
+    )
```
fengyuentau, line 96: How about we keep it simple with scale=1? I guess the jittering face boxes come from multi-scale detection.

models/facial_expression_recognition/demo.py
```diff
@@ -0,0 +1,163 @@
...
+
+    return dets, fer_res
+
+
+if __name__ == '__main__':
+    detect_model = YuNet(modelPath=args.detect_model,
+                         inputSize=[320, 320],
+                         confThreshold=args.conf_threshold,
+                         nmsThreshold=args.nms_threshold,
+                         topK=args.top_k,
```
fengyuentau, line 113: use default values for `modelPath`, `confThreshold`, `nmsThreshold` and `topK` to save arguments.

models/facial_expression_recognition/demo.py
```diff
@@ -0,0 +1,163 @@
...
+    if args.input is not None:
+        image = cv.imread(args.input)
+
+        # Get detection and fer results
+        dets, fer_res = process(detect_model, align_model, fer_model, image)
+
+        # Draw results on the input image
+        image = visualize(image, dets, fer_res)
+
+        print('%s %3d faces detected.' % (datetime.datetime.now(), len(dets)))
```
fengyuentau, line 133: Some times the text on face bbox is not clear enough to recognize. How about moving the print into `visualize` and  print faces with coordinates and facial expression?

fengyuentau (commented):
Thank you for your contribution 👍! There are some minor comments below. Other looks good to me. I will benchmark the model on the hardware farm on my side. Could you provide two or three images (each has only one face) for benchmarking?

By the way, I think we can have a high level API dedicated for this model in the objdetect module.

models/facial_expression_recognition/quantize-inc.py
```diff
@@ -0,0 +1,125 @@
+import os
```
fengyuentau, line 1: I think this file is basically the same with the one in tools/quantize except the different default scale used in `Dataset`. Could you update `tools/quantize/quantize-inc.py` with the changes in this file and remove this file?

Others look good to me. I can help you rebase commits so that we can merge this pull request.

fengyuentau (approved):

fengyuentau (approved):
👍

PR #106, opened 2022-11-13
Data and benchmarking for the new object detection models

fengyuentau:
TODO:
- [x] upload data to gdrive and baidu pan
- [x] add link in download_data.py
- [x] add samples in the front readme
- [x] add notes in each model's readme on where the original sample images can be fetched
- [x] benchmark configs for nanodet and yolox
- [x] update benchmark data for nanodet and yolox

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,13 @@ Guidelines:
 
 ## Models & Benchmark Results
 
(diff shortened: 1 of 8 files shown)


fengyuentau:
👍

PR #110, opened 2022-12-03
Fix variable name error in yolox.py/setBackend

ruoningYu:
Fix variable name error in yolox.py/setBackend


fengyuentau (approved):
Thank you for the contribution! Looks good to me 👍

Issue #122, opened 2023-01-25
Failed to load CRNN FP16 models with OpenCV 4.7 


fengyuentau:
Reproducer:

```python
import cv2 as cv

cv.dnn.readNet("text_recognition_CRNN_CH_2022oct_fp16.onnx")
cv.dnn.readNet("text_recognition_CRNN_EN_2022oct_fp16.onnx")
```

Error log:

```
[ WARN:0@0.032] global onnx_graph_simplifier.cpp:804 getMatFromTensor DNN: load FP16 model as FP32 model, and it takes twice the FP16 RAM requirement.
[ERROR:0@0.049] global onnx_importer.cpp:1054 handleNode DNN/ONNX: ERROR during processing node with 2 inputs and 1 outputs: [Conv]:(onnx_node!Conv_0) from domain='ai.onnx'
Traceback (most recent call last):
  File "/.../opencv_zoo/models/text_recognition_crnn/demo.py", line 63, in <module>
    recognizer = CRNN(modelPath=args.model)
  File "/.../opencv_zoo/models/text_recognition_crnn/crnn.py", line 16, in __init__
    self._model = cv.dnn.readNet(self._model_path)
cv2.error: OpenCV(4.7.0) /Users/opencv-cn/GHA-OCV-1/_work/opencv-python/opencv-python/opencv/modules/dnn/src/onnx/onnx_importer.cpp:1073: error: (-2:Unspecified error) in function 'handleNode'
> Node [Conv@ai.onnx]:(onnx_node!Conv_0) parse error: OpenCV(4.7.0) /Users/opencv-cn/GHA-OCV-1/_work/opencv-python/opencv-python/opencv/modules/dnn/src/layers/convolution_layer.cpp:398: error: (-215:Assertion failed) inputs.size() != 0 in function 'getMemoryShapes'
>

[ WARN:0@0.024] global onnx_graph_simplifier.cpp:804 getMatFromTensor DNN: load FP16 model as FP32 model, and it takes twice the FP16 RAM requirement.
[ERROR:0@0.027] global onnx_importer.cpp:1054 handleNode DNN/ONNX: ERROR during processing node with 3 inputs and 1 outputs: [Conv]:(onnx_node!Conv_0) from domain='ai.onnx'
Traceback (most recent call last):
  File "/.../opencv_zoo/models/text_recognition_crnn/demo.py", line 63, in <module>
    recognizer = CRNN(modelPath=args.model)
[...]

PR #123, opened 2023-01-25
Add model file collector, fix some bugs and add some features

### LINKED ISSUE #120: Add a model file path collector in `models/__init__.py`
We now have lots of configs for benchmark and each of them has a hard written model path like the following:

```yaml
Model:
  name: "YuNet"
  modelPath: "models/face_detection_yunet/face_detection_yunet_2022mar.onnx"
```

This is not convenient and bad for automation when it comes to upgrade a model and benchmark int8-quantized model. To address this problem, Add a model file path collector `models/__init__.py` and also add an option in benchmark configs to allow loading int8-quantized models.

fengyuentau:
Resolves #120 

This pull request:
- adds the model file searcher and removes the hard-written file paths from the benchmark configs,
- packs ppresnet, mobilenet and crnn with labels, and therefore hard-written file paths are removed,
- fixes the palm det data file which was not organized in a directory previously,
- excludes the benchmark of fp16 crnn models due to https://github.com/opencv/opencv_zoo/issues/122

To do feature:
- [Done] add a flag in benchmark.py to control running a kind of model only, such as fp32, fp16, int8, ...
- [Yes, but no plan to do so] can tools/quantize use the collector as well?
    - Hard-written file paths work great if anyone wants to know which model is quantizing.
- [Yes, but no plan to do so] can tools/eval use the collector as well?
    - Same above.


fengyuentau:
👍

PR #128, opened 2023-02-07
Update palm detection model from MediaPipe (2023feb)

WanliZhong:
This PR aims to change the palm detection model from tfjs model to mediapipe model. Model downloaded from https://google.github.io/mediapipe/solutions/models.html#hands

The main updates between the previous mdoel and the current model are as follows.

**Model Information:**
|              | Previous Model 	| Current Model 	|
|:--------------:	|:--------------:	|:-------------:	|
|   Model Size   	|     6.78MB     	|     3.72MB    	|
|   Input Size   	|    256\*256    	|    192\*192   	|

**Inference Time:** 
|    Platform    	| Previous Model 	| Current Model 	|
|:--------------:	|:--------------:	|:-------------:	|
|    Intel-CPU   	|     15.57ms    	|    11.09ms    	|
|     RPI-CPU    	|    168.37ms    	|    63.79ms    	|
|   JETSON-GPU  	|     50.64ms    	|    83.20ms    	|
|    KV3-NPU   	|     62.45ms    	|    33.81ms    	|

**Model Accuracy**
Current model has better performance by running demo. Also fixed the bug that the aspect ratio was not kept when resizing images.
| Previous Model 	| Current Model 	|
|:--------------:	|:-------------:	|
|| |


WanliZhong:
The accuracy of the `int8` quantization model decreases more, but the speed improves a lot. The following video is a demo of the `int8` model.

https://user-images.githubusercontent.com/49567307/217290047-56550fc3-51c6-4294-80c9-ecd0dc1ecc2f.mp4

ShiqiYu:
What's the main difference between this one and the previous one?

WanliZhong:
> What's the main difference between this one and the previous one?

Thanks for reminding! I've updated the description of this PR.

README.md
```diff
@@ -34,7 +34,7 @@ Guidelines:
 | [WeChatQRCode](./models/qrcode_wechatqrcode)            | QR Code Detection and Parsing | 100x100    | 7.04           | 37.68        | ---             | ---          | ---         |
 | [DaSiamRPN](./models/object_tracking_dasiamrpn)         | Object Tracking               | 1280x720   | 36.15          | 705.48       | 76.82           | ---          | ---         |
 | [YoutuReID](./models/person_reid_youtureid)             | Person Re-Identification      | 128x256    | 35.81          | 521.98       | 90.07           | 44.61        | ---         |
-| [MP-PalmDet](./models/palm_detection_mediapipe)         | Palm Detection                | 256x256    | 15.57          | 168.37       | 50.64           | 62.45        | ---         |
+| [MP-PalmDet](./models/palm_detection_mediapipe)         | Palm Detection                | 192x192    | 11.09          | 63.79        | 83.20           | 33.81        | ---         |
```
fengyuentau, line 37: Slower on Jetson?

WanliZhong: After verification, this problem is caused by OpenCV version iteration. In version `4.6.0`, time cost on cuda should change **from 50ms to 30ms**. But in version `4.7.0`, time cost on cuda change **from 200ms to 80ms**.

This means that after the OpenCV version update, inference time of this palm detecion model is **4 times slower** on cuda. Let's solved this problem firstly.

ShiqiYu:
I tested the hand pose demo on Khadas VIM3. The FPS is the almost the same whatever I choose the TIMVX+NPU or Default+CPU. Could you check if everything is right?

ShiqiYu:
The accuracy of the INT8 palm detector is obviously lower than that of the float model. Should we release the int8 model?

WanliZhong:
HandPose model still need to be updated after this PR.

New palm detection model on Vim3 do have the same spped no matter if we use npu. I'm not sure why this happen.

I think we can add more images to quantized this model so that the accuracy may be higher.

I will try to improve this model.

fengyuentau:
I think we can include @zihaomu for discussion on the speed.

zihaomu:
image

> New palm detection model on Vim3 do have the same spped no matter if we use npu. I'm not sure why this happen.

I have compared the new and old model. And the new model has lot of `PRelu` layer. Because `PeRelu` is float32, it will cause the model to be cut into many small graph when NPU is executed. Every switch from float to int8 will be very time consuming, that's why you won't get NPU acceleration.

WanliZhong:
I re-quantified the model with 3900+ images and got a better model. The accuracy is obviously better, but the inference time is longer on CPU.

image

README.md
fengyuentau: Leave as-is and will be fixed in the next release.

WanliZhong: This problem is fixed by PR https://github.com/opencv/opencv/pull/23243.

models/palm_detection_mediapipe/README.md
```diff
@@ -1,10 +1,11 @@
 # Palm detector from MediaPipe Handpose
 
-This model detects palm bounding boxes and palm landmarks, and is converted from Tensorflow-JS to ONNX using following tools:
+This model detects palm bounding boxes and palm landmarks, and is converted from TFLite to ONNX using following tools:
```
fengyuentau, line 3: By the way, TFLite support in DNN has been merged. Dev branch can now load and run TFLite models.

fengyuentau (approved):
👍

PR #129, opened 2023-02-08
Update FER quantized model and fix document

WanliZhong:
This PR try to quantize `FER` model in `per_tensor` mode. Since the accuracy evaluation dataset has 350 images, the quantize loss should be relaxed. I changed it from `0.01` to `0.02` and then got a acceptable quantized model. 

The accuracy of the current model in the quantize evaluation dataset is `int8|fp32: 0.7457|0.7571`, which is only droped a little. But the time cost in `KV3-NPU` is changed **from 108.53ms to 29.80ms**!


fengyuentau (approved):
👍

Issue #119, opened 2023-01-15
Add evaluation scripts for models in the zoo


fengyuentau:
We now have over 15 models covering more than 10 tasks in the zoo. Although most of the models are converted to ONNX straightly from its original format, such conversion may potentially lead to drop of accuracy, especially for FP16 and Int8-quantized models. To show the actual accuracy for our users, we now already have some evaluation scripts with the following conditions met in https://github.com/opencv/opencv_zoo/tree/master/tools/eval:

1. **Reproduce the claimed accuracy** with the converted FP32 ONNX model using OpenCV DNN as inference framework. The claimed accuracy is either from the source repository or paper on the same dataset, which needs to be specified in the first comment of the pull request.
2. Once it is reproduced, **apply the same evaluation script** on FP16 and Int8-quantized models.

Take a look at the task list below for current status. **Feel free to leave a comment for application or discussion before you start to contribute.**

| Status | Task | Dataset | Models |
| - | - | - | - |
| ✅ Done  | Face Detection | WIDERFace | YuNet |
| ✅ Done  | Face Recognition | LFW | SFace |
| ❗️ Need Contribution | License Plate Detection | ? | LPD-YuNet |
| 🏃 https://github.com/opencv/opencv_zoo/pull/91 | Object Detection | COCO | YOLOX & NanoDet |
| 🏃 https://github.com/opencv/opencv_zoo/pull/73 | Text Detection | ? | DB |
| ✅ Done | Text Recognition | ICDAR2003 & IIIT5K | CRNN (EN & CN) |
| ✅ Done | Image Classification | ImageNet | PP-ResNet50 & MobileNet V1 / V2 |
| ❗️ Need Contribution | Human Segmentation | ? | PP-HumanSeg |
| ❗️ Need Contribution | QR Code Detection / Parsing | ? | WeChatQRCode |
| ❗️ Need Contribution | Object Tracking | ? | DaSiamRPN |
| ❗️ Need Contribution | Person Re-identification | ? | YoutuReID |
| ❗️ Need Contribution | Palm Detection | ? | MP-PalmDet |
| ❗️ Need Contribution | Hand Pose Estimation | ? | MP-HandPose |

labeeb-7z:
hey @fengyuentau , i'd love to contribute to this and need some clarification.

say if I were to add eval script for the PP-HumanSeg model, I should
 -  add a coco.py file in `tools/eval/datasets` 
 - this file will implement a coco class with the following methods.
[...]

PR #130, opened 2023-02-10
added evaluation script for PPHumanSeg model

labeeb-7z:
This PR is meant to add an evaluation script for the PP Human segmentation model present in the zoo as mentioned in this issue #119 .

I referred to the val script used by PaddleSeg which can be found [here](https://github.com/PaddlePaddle/PaddleSeg/blob/release/2.1/contrib/HumanSeg/val.py) and [here](https://github.com/PaddlePaddle/PaddleSeg/blob/release/2.1/paddleseg/core/val.py) and implemented the evaluation script for PPHumanSeg in the method other evaluation scripts on zoo are written.

This is the ouput I'm getting after running the evaluation script : 

Screenshot from 2023-02-10 09-55-30

Let me know if this PR can be improved further. Thanks!


tools/eval/README.md
```diff
@@ -181,7 +182,7 @@ python eval.py -m crnn -d icdar -dr /path/to/icdar
 
 ### Prepare data
 
-Please visit https://github.com/cv-small-snails/Text-Recognition-Material to download the IIIT5K dataset and the labels.
+Please visit https://github.com/cv-small-snails/Text-Recognition-Material to download the IIIT5K dataset and the labels.
```
fengyuentau, line 185: Extra whitespace

tools/eval/README.md
```diff
@@ -190,4 +191,18 @@ Run evaluation with the following command:
 
 ```shell
 python eval.py -m crnn -d iiit5k -dr /path/to/iiit5k
+```
+
+
+## PPHumanSeg
```
fengyuentau, line 197: I wonder if we can use the `mini_supervisely` for the name of dataset so as to avoid confusion with the model itself.

tools/eval/datasets/pp_humanseg.py
```diff
@@ -0,0 +1,209 @@
+import os
+import cv2 as cv
+import numpy as np
+from tqdm import tqdm
+
+
+class PPHumanSeg :
```
fengyuentau, line 7: Propose to change the name to `MiniSupervisely`

tools/eval/datasets/pp_humanseg.py
```diff
@@ -0,0 +1,209 @@
...
+        print("Class IoU : ", self.class_iou)
+        print("Mean Accuracy : ", self.acc)
+        print("Class Accuracy : ", self.class_acc)
+
+    
+    def one_hot(self, arr, max_size) :
+        return np.eye(max_size)[arr]
+
+
+    def calculate_area(self,pred, label, num_classes, ignore_index=255):
```
fengyuentau, line 109: Could you provide some test data for validating the correctness of this function?

tools/eval/datasets/pp_humanseg.py
```diff
@@ -0,0 +1,209 @@
...
+            label_area_i = np.sum(label_i)
+            intersect_area_i = np.sum(pred_i * label_i)
+            pred_area.append(pred_area_i)
+            label_area.append(label_area_i)
+            intersect_area.append(intersect_area_i)
+        
+        return intersect_area, pred_area, label_area
+    
+
+    def mean_iou(self,intersect_area, pred_area, label_area):
```
fengyuentau, line 155: Could you provide some test data for validating the correctness of this function? Something like test cases with given input and expected output.

tools/eval/eval.py
```diff
@@ -97,6 +101,9 @@
         iiit5k=dict(
             name="IIIT5K",
             topic="text_recognition"),
+        pphumanseg=dict(
+            name="PPHumanSeg",
+            topic="human_segmentation"),
```
fengyuentau, line 106: Propose to change dataset name.

tools/eval/eval.py
```diff
@@ -108,7 +115,6 @@ def main(args):
     model_topic = models[model_key].pop("topic")
     model_handler, _ = MODELS.get(model_name)
     model = model_handler(**models[model_key])
-
```
fengyuentau, line 111: unnecessary removal

fengyuentau (commented):
Thank you for contribution! Please, take a look at the comments below.

WanliZhong:
@labeeb-7z Thanks for your contribution! Could you add the evaluation results of `fp32` and `int8` models on [PPhumanseg Readme](https://github.com/opencv/opencv_zoo/tree/master/models/human_segmentation_pphumanseg#pphumanseg)? You can add the information like the example below:

-------------------------------------------------------------------
Results of accuracy evaluation with [tools/eval](../../tools/eval).

| Models      | Mean Accuracy |
| ----------- | -------- |
| PPHumanSeg| 0.9023   |
| PPHumanSeg quant | 0.xxxx   |

\*: 'quant' stands for 'quantized'.

tools/eval/datasets/pp_humanseg.py
labeeb-7z: For the following test : 
``` 
# tensor of size (1,x,x)
pred = np.array([
    [   [0, 0, 0, 0, 0],
        [0, 0, 1, 1, 0],
        [0, 0, 1, 1, 0],
        [0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0]],
])
label = np.array([
    [   [0, 0, 0, 0, 0],
        [0, 0, 1, 1, 0],
        [0, 0, 1, 1, 0],
        [0, 0, 0, 1, 0],
        [0, 0, 0, 0, 0]],
])

area = calculate_area(pred, label, 2)
print(f"intersect_area: {area[0]}  pred_area: {area[1]}  label_area: {area[2]}")
iou = mean_iou(*area)
print(f"IOU on classes: {iou[0]}  mean IOU: {iou[1]}")
```

Here's the output : 
image

It can be seen from the pred and label tensor that intersect_area of 0s is 20 and intersect area of 1s is 4.

For IOU : 
union for class 0 : pred + label - interset = 20 + 21 - 20 = 21
iou for class 0 : intersect/union = 20/21 = 0.9523
union for class 1 : pred + label - interset  4 + 5 - 4 = 5
iou for class 1 : intersect/union = 4/5 = 0.8

tools/eval/datasets/pp_humanseg.py
labeeb-7z: please refer above

labeeb-7z:
@WanliZhong here's the output I got for quantized model : 
image

WanliZhong:
> @WanliZhong here's the output I got for quantized model : image

There is too much loss of accuracy in the quantized model and we need to consider re-quantizing the model. Thank you for the evaluation script! 👍

models/human_segmentation_pphumanseg/README.md
```diff
@@ -22,6 +22,18 @@ python demo.py --help
 
 ![messi](./examples/messi.jpg)
 
+---
+Results of accuracy evaluation with [tools/eval](../../tools/eval).
+
+| Models             | Accuracy | 
+| ------------------ | -------------- |
+| PPHumanSeg         | 0.9023         |
+| PPHumanSeg quant   | 0.4940         |
```
fengyuentau, line 31: Column `Accuracy` is not fully aligned.

tools/eval/eval.py
```diff
@@ -108,7 +119,7 @@ def main(args):
...
+
```
fengyuentau, line 122: Extra whitespaces here

fengyuentau (commented):
Some minor comments below. I will have a check on the implementation of the evaluation metric later.

labeeb-7z:
For more reference on the evaluation metric you can have a look at the [metrics module](https://github.com/PaddlePaddle/PaddleSeg/blob/release/2.1/paddleseg/utils/metrics.py) of PaddleSeg which is used by the evaluation script of PaddleSeg I linked in PR description.

fengyuentau (commented):
I ran the official evaluation script and the mIoU of the model should be 0.8602. I found there are many differences between your implementation and the official evaluation script. Below are some differences I found.

I am not saying you have to follow the official implementation completely but just make sure it is correct. If you cannot ensure correctness, it will save a lot of time.

labeeb-7z:
I've moved the *_all variables outside the loop (apologies for that). As for ignoring index and adding 1, I've tried to followed the PaddleSeg implementation. And for the one_hot vectors, again I have implemented the one_hot function of PaddleSeg and followed their evaluation script.

Let me know of any other changes. The updated miou is : 
image

tools/eval/datasets/minisupervisely.py
```diff
@@ -0,0 +1,206 @@
...
+
+        intersect_area_all = []
+        pred_area_all = []
+        label_area_all = []
+
+        pbar = tqdm(self.image_set)
+
+        for input_image, expected_image in pbar : 
+            pbar.set_description(
+                "Evaluating {} with {} val set".format(model.name, self.name))
```
fengyuentau, line 56: These two lines should be placed outside the for loop since the description does not need to be updated every single iteration.

tools/eval/datasets/minisupervisely.py
```diff
@@ -0,0 +1,206 @@
...
+            
+            input_image = cv.imread(os.path.join(self.root, input_image))
+            input_image = cv.resize(input_image, (192, 192))
```
fengyuentau, line 59: Missing two transforms:

1. to_rgb: https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/paddleseg/transforms/transforms.py#L61-L62. Note that to_rgb is set to True by default.
2. normailze: 
    1. Config: https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/contrib/HumanSeg/configs/fcn_hrnetw18_small_v1_humanseg_192x192_mini_supervisely.yml#L20-L28 
    2. Normalize transform: https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/paddleseg/transforms/transforms.py#L402-L444

fengyuentau (commented):
Thank you for the update! Lets try to reproduce the exact mIoU number.

tools/eval/datasets/minisupervisely.py
labeeb-7z: I dont see a direct transforms applied in the [infer method](https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/paddleseg/core/infer.py#L192) used by the [evaluation method](https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/paddleseg/core/val.py#L28).  The [inference script](https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/legacy/contrib/HumanSeg/infer.py#L68) of the model in PaddleSeg uses it. So Im not sure if there's a need for it in this eval script

tools/eval/datasets/minisupervisely.py
labeeb-7z: This was referred from [eval script of icdar](https://github.com/opencv/opencv_zoo/blob/6c68bc48c6f96042b29b3425174e431ccac38376/tools/eval/datasets/icdar.py#L37).
I've changed it for this script, should I also change it for icdar?

fengyuentau: > I've changed it for this script, should I also change it for icdar?

Not in this pull request.

tools/eval/datasets/minisupervisely.py
fengyuentau: > mIoU has decreased to 0.3562 .

My suggestion would be doing alignment with PaddleSeg's results image by image. For example, you could add `break` right after `calculate_area` in your code and PaddleSeg, print `intersect_area, pred_area, label_area` to see whether your results are completely the same with PaddleSeg's. You will also need to ensure you are comparing the same image since PaddleSeg uses a data loader to load images which may have shuffled images.

labeeb-7z: @fengyuentau The reason mIoU decresead to 0.3562 was doing normalization twice. [model.infer](https://github.com/opencv/opencv_zoo/blob/6c68bc48c6f96042b29b3425174e431ccac38376/models/human_segmentation_pphumanseg/pphumanseg.py#L44) of PPHumanSeg already does normalization before passing the input to model so we dont have to do normalization again in the evaluation script. I've reverted that back.

The reason behind why current eval script gives mIoU of 0.8798 and not the expected 0.8602 is - model outputs a 192x192 tensor(logits). Original PaddleSeg script resizes this tensor back to original image size before taking one-hot of that tensor and calculating accuracy, IoU (all calculations done on tensor of shape equal to input image).
This can be seen below : 
[infer method called by PaddleSeg evaluation script](https://github.com/PaddlePaddle/PaddleSeg/blob/13aac35b31b147d2dd9976821ade13dabebff685/paddleseg/core/infer.py#L134)
image
Here's the output (the input image is of size 800x1313 in this case) : 
image

On the other hande, current eval script resizes the expected_image (label) to 192x192 (same as output of model) and then calculates accuracy, IoU. 
This is because the infer method of PPHumanSeg does not resize the input back to its original shape and directly calculates the one-hot. And thus all the calculation (for mIoU and accuracy) is done on a 192x192 tensor and not the original image size tensor.
image

image


Should I proceed with modifying the infer method of PPHumanSeg class to resize the output back to original image shape before calculating the one-hot?

fengyuentau: > Should I proceed with modifying the infer method of PPHumanSeg class to resize the output back to original image shape before calculating the one-hot?

No, the idea of infer method of PPHumanSeg is assuming the input image is resized to the target size [192, 192] already. Thus, any resize operation should be done here, say right after `output_image = model.infer(input_image)   ` we `reverse_transform(output_image)`.

`to_rgb` needs to be kept or directly add this in `PPHumanSeg._preprocess`.

labeeb-7z: The logits output of model.infer should be resized. model.infer returns the one-hot and not the logits.

so `output_image` is a one-hot and resizing that won't be the same as resizing the `logits` of the model.(correct me if Im wrong).

If we want to do `reverse_transform(outpu_image)` in the eval script, we should change the model.infer method to return the logits and not postprocess it to get the one-hot.

So
- resize the logits in model.infer method before calculating one-hot(postprocess)
- return the logits from model.infer and do reverse_transform in the eval scrirpt.

which one should we proceed with?

fengyuentau: Okay. I propose to add resize (to target_size [192, 192]) in `_preprocess` and reverse resize (to its original size) in `_postprocess` in `PPHumanSeg` instead.

labeeb-7z: I made the above changes, however the mIoU now is 0.9085 .

I started by comparing the 2 models. The [human_segmentation_pphumanseg_2021oct.onnx](https://github.com/opencv/opencv_zoo/blob/master/models/human_segmentation_pphumanseg/human_segmentation_pphumanseg_2021oct.onnx) downloaded from [this repository](https://github.com/opencv/opencv_zoo/tree/master/models/human_segmentation_pphumanseg) was evaluated on the script I've written so far. 
The [PP-HumanSegV1-Lite](https://github.com/PaddlePaddle/PaddleSeg/tree/release/2.7/contrib/PP-HumanSeg#42-general-human-segmentation-models) model downloaded from PaddleSeg repo (branch release/2.7). It has mUoU of 0.8602 . It was evaluated using the [eval script](https://github.com/PaddlePaddle/PaddleSeg/tree/release/2.7/contrib/PP-HumanSeg#63-evaluation) provided by PaddleSeg

First I compared the input to both models (make sure preprocessing is done properly) : 
zoo model : 
image
PaddleSeg model ([infer method](https://github.com/PaddlePaddle/PaddleSeg/blob/83a419ff92c54b01257c75d75a435df6d75e7515/paddleseg/core/infer.py#L192) called by eval script): 
image

Both the tensors were equivalent : 
image


Next I compared the immediate outputs of the model : 
zoo model : 
Screenshot from 2023-02-16 23-15-02
PaddleSeg model : 
image

Surprisingly these 2 outputs were different!
Screenshot from 2023-02-16 23-15-23

(I made sure both of them were using the same input image).

Most likely Im using a different model for comparision? Let me know where can I find the actual model or if Im doing something else wrong

fengyuentau: > Most likely Im using a different model for comparision?

It could happen. The one here in the zoo is converted from Paddle to ONNX using [this script](https://github.com/opencv/opencv/tree/master/samples/dnn/dnn_model_runner/dnn_conversion/paddlepaddle#2-run-paddleseg-portrait-segmentation-using-opencv), and the script has been there for 2 years. Note that the model was provided in a link which is different from all other links in PaddleSeg:release/2.7. Paddle could have updated this model during this time and somehow still name it "v1".

It can be verified with using your evaluation script to evaluate the one from PaddleSeg:release/2.7. Paddle provides a conversion tool https://github.com/PaddlePaddle/Paddle2ONNX. If you can reach the same mIoU (0.8602) with your evaluation script and the model from PaddleSeg:release/2.7, that's a success.

labeeb-7z: I tried to convert the models from [PaddleSeg](https://github.com/PaddlePaddle/PaddleSeg/tree/release/2.7/contrib/PP-HumanSeg#42-general-human-segmentation-models) into onnx format.

the PP-HumanSegV1-Lite (checkpoint) gives the following error
image

I tried other models (v2,mobile) from there, none of the checkpoint models were converted to onnx (same error).

then I tried the Inference Model(softmax) of v1. It was converted to onnx successfully.
However, `cv.dnn.readNet()` wasn't able to read it : 
image

The readNet function gave the same error for Mobile inference model(softmax).

At this point, PaddleSeg doesn't provide the model currently in this repository, and the models they provide either don't convert to onnx or are not readable by readNet().

What should be the next steps?

fengyuentau: > However, cv.dnn.readNet() wasn't able to read it 

The error was due to lack of support of dynamic input shape (dynamic H or W) I believe. You can use https://github.com/daquexian/onnx-simplifier to simplify it with a fixed input shape, such as [1, 3, 192, 192]. After that, OpenCV DNN should be able to load and infer the model.

fengyuentau: > However, cv.dnn.readNet() wasn't able to read it 

The error was due to lack of support of dynamic input shape (dynamic H or W) I believe. You can use https://github.com/daquexian/onnx-simplifier to simplify it with a fixed input shape, such as [1, 3, 192, 192]. After that, OpenCV DNN should be able to load and infer the model.

labeeb-7z: I converted the v1 model using the command 
`onnxsim seg.onnx v1sim.onnx --overwrite-input-shape 1,3,192,192`

but now the onnx model is corrupted in some way : 
Screenshot from 2023-02-17 15-32-43

fengyuentau: > I tried the Inference Model(softmax) of v1. It was converted to onnx successfully.

Did it really work? I downloaded the same model and use the same command as yours but the produced model has 0 Bytes.

labeeb-7z: On [PaddleSeg](https://github.com/PaddlePaddle/PaddleSeg/tree/release/2.7/contrib/PP-HumanSeg#42-general-human-segmentation-models) for PP-HumanSegV1-Lite, the model downloaded with `checkpoint` link gives the error and produces a model of 0 Bytes.

But the model downloaded with `Inference Model (Softmax)` link is converted to onnx succesfully : 
image
image

I simplified the onnx  models produced by `Inference Model (Softmax)` and than tried the eval script on them.
These models were not read by `cv.dnn.readNet` which I shared above

fengyuentau: Thank you for pointing out the correct model and command. The conversion was done successfully and the simplifier worked as well. Latest OpenCV (4.7) can load and infer the model without any issues. See my screen shot below. You could have a try with my model: [humanseg.sim.onnx.zip](https://github.com/opencv/opencv_zoo/files/10779034/humanseg.sim.onnx.zip)


image

tools/eval/datasets/minisupervisely.py
```diff
@@ -0,0 +1,201 @@
...
+        Print evaluation results
+        """
+        print("Mean IoU : ", self.miou)
+        print("Mean Accuracy : ", self.acc)
+
+    
+    def one_hot(self, arr, max_size) :
+        return np.eye(max_size)[arr]
+
+
```
fengyuentau, line 105: Looks like it is not used anymore. Suggest to remove this.

tools/eval/datasets/__init__.py
```diff
@@ -20,4 +21,5 @@ def register(self, item):
 DATASETS.register(WIDERFace)
 DATASETS.register(LFW)
 DATASETS.register(ICDAR)
-DATASETS.register(IIIT5K)
\ No newline at end of file
+DATASETS.register(IIIT5K)
+DATASETS.register(MiniSupervisely)
```
fengyuentau, line 25: nit: put a newline at the end of the file.

tools/eval/datasets/minisupervisely.py
labeeb-7z: Thanks, the error was fixed by removing the `self.outputNames` argument from `model.forward()`.

The eval script achieved the expected mIoU of 0.8602 on the model you shared above  (humanseg.sim.onnx) : 
image

This matches the results from the PaddleSeg script : 
image
 
With the same script, for the model present in the zoo, the mIoU is 0.8996 : 
image

fengyuentau: 👍

tools/eval/README.md
```diff
@@ -190,4 +191,24 @@ Run evaluation with the following command:
 
 ```shell
 python eval.py -m crnn -d iiit5k -dr /path/to/iiit5k
+```
+
+
+## mini_supervisely
```
fengyuentau, line 197: -> Mini Supervisely

tools/eval/README.md
```diff
@@ -190,4 +191,24 @@ Run evaluation with the following command:
...
+
+### Prepare data
+Please download the mini_supervisely data from [here](https://paddleseg.bj.bcebos.com/humanseg/data/mini_supervisely.zip) which includes the validation dataset and unzip it.
+
+### Evaluation
+
+Run evaluation with the following command
```
fengyuentau, line 204: add a colon at the end of this line

tools/eval/README.md
```diff
@@ -190,4 +191,24 @@ Run evaluation with the following command:
...
+
+```shell
+python eval.py -m pphumanseg -d mini_supervisely -dr /path/to/pphumanseg
+```
+
+Run evaluation on quantized model with the following command
```
fengyuentau, line 210: add a colon at the end of this line

tools/eval/datasets/minisupervisely.py
```diff
@@ -0,0 +1,196 @@
+import os
+import cv2 as cv
+import numpy as np
+from tqdm import tqdm
+
+
+class MiniSupervisely :
```
fengyuentau, line 7: Could you add a comment as reference to the original evaluation method in PaddleSeg? Something like this:

```.diff
class MiniSupervisely : 
+    '''
+       Refer to https://github.com/PaddlePaddle/PaddleSeg/blob/release/2.7/paddleseg/core/val.py
+       for official evaluation implementation.
+    '''
```

fengyuentau (commented):
Overall looks good 👍 We can merge this pull request after the comments below are resolved.

labeeb-7z:
@fengyuentau I've made the suggested changes, Thanks!

fengyuentau (approved):
Thank you for contribution! 👍

PR #134, opened 2023-02-21
move setting description outside for loop in evaluation scripts

labeeb-7z:
This PR moves the `pbar.set_description` line outside of the for loop as discussed [here](https://github.com/opencv/opencv_zoo/pull/130#discussion_r1105252865) in  the following evaluation scripts : 
- [icdar.py](https://github.com/opencv/opencv_zoo/blob/6c68bc48c6f96042b29b3425174e431ccac38376/tools/eval/datasets/icdar.py#L37)
- [imagenet.py](https://github.com/opencv/opencv_zoo/blob/6c68bc48c6f96042b29b3425174e431ccac38376/tools/eval/datasets/imagenet.py#L43)
- [iiitk5.py](https://github.com/opencv/opencv_zoo/blob/6c68bc48c6f96042b29b3425174e431ccac38376/tools/eval/datasets/iiit5k.py#L40)

### DIFF
--- a/tools/eval/datasets/icdar.py
+++ b/tools/eval/datasets/icdar.py
@@ -33,8 +33,9 @@ class ICDAR:
     def eval(self, model):
         right_num = 0
         pbar = tqdm(self.val_label)
+        pbar.set_description("Evaluating {} with {} val set".format(model.name, self.name))
+
         for fn, label in pbar:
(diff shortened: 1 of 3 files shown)


fengyuentau (approved):
Thank you for the fix! 👍

Issue #136, opened 2023-02-23
Some models do not work with CANN backend


fengyuentau:
List of models not working with CANN backend:

- [x] lpd_yunet
- [x] nanodet
- [x] mobilenet v2
- [x] pphumanseg
- [ ] dasiamrpn, cannot reproduce without the API, loading separately is good, not going to fix it.
- [x] handpose mp

Fixed by https://github.com/opencv/opencv/pull/23319 except DaSiamRPN.

The following models need to be updated:

- lpd_yunet, bump the opset version of Slice.
- pphumanseg, replace with an newer and simpler one.

Reproducer of lpd_yunet:

```shell
$ python3 benchmark.py --cfg config/license_plate_detection_yunet.yaml --fp32
Benchmarking LPD_YuNet with ['license_plate_detection_lpd_yunet_2022may.onnx']
Traceback (most recent call last):
  File "benchmark.py", line 143, in <module>
    benchmark.run(model)
  File "benchmark.py", line 103, in run
    self._benchmark_results[filename][str(size)] = self._metric.forward(model, *data[1:])
  File "/home/test_user01/fytao/opencv_zoo/benchmark/utils/metrics/detection.py", line 21, in forward
    model.infer(img)
  File "/home/test_user01/fytao/opencv_zoo/models/license_plate_detection_yunet/lpd_yunet.py", line 56, in infer
    outputBlob = self.model.forward(self.output_names)
cv2.error: OpenCV(4.7.0-dev) /home/test_user01/fytao/opencv-opencv/modules/dnn/src/layers/slice_layer.cpp:640: error: (-215:Assertion failed) sliceSteps.size() == 1 in function 'initCann'
```

Looks like the issue is the out-of-date Slice in the model (`starts` and `ends` are attributes, and no `steps`). Should have a try with upgrading opset version of lpd_yunet.

fengyuentau:
NanoDet does not work as well with CANN backend:

```shell
$ python3 benchmark.py --cfg config/object_detection_nanodet.yaml --fp32
Benchmarking NanoDet with ['object_detection_nanodet_2022nov.onnx']
Traceback (most recent call last):
  File "benchmark.py", line 143, in <module>
    benchmark.run(model)
  File "benchmark.py", line 103, in run
    self._benchmark_results[filename][str(size)] = self._metric.forward(model, *data[1:])
  File "/home/test_user01/fytao/opencv_zoo/benchmark/utils/metrics/detection.py", line 21, in forward
[...]

PR #137, opened 2023-02-23
Update benchmark and results for CANN backend

fengyuentau:
Known issues:
- https://github.com/opencv/opencv_zoo/issues/136, almost fixed by https://github.com/opencv/opencv/pull/23319
- Integrated APIs such as FaceDetectorYN, setting backend and target separately fails. Possible solution is make an API in those wrapper classes like `setBackendAndTarget(backend_id, target_id)` in place of `setBackend` and `setTarget`.

Next PR:
- Limit the combination of backends and targets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,27 +15,27 @@ Guidelines:
 
 ## Models & Benchmark Results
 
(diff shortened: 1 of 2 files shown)


WanliZhong (approved):
👍👍👍The results of CANN backend is really amazing!

PR #141, opened 2023-03-08
Update hadnpose preprocess

WanliZhong:
This PR try to fix conner cases shown below: after rotateing hand, there are some pixels may affect the result will be lost.
image

By the way, pre-croping the image to a range of interest (RoI) may accelerate preprocess when original image is too large.


WanliZhong:
Benchmark cannot test the speed difference because the images in the dataset are small. There is no significant change when running the demo.

fengyuentau (approved):
👍

PR #145, opened 2023-03-16
Limit combinations of backends and targets in demos and benchmark

### LINKED ISSUE #144: Limit the combinations of targets and backends
Currently the combinations of targets and backends are too random and actually some combinations like `DNN_BACKEND_CANN` and `DNN_TARGET_CPU` are not valid. So I propose to do the following changes:

- In demos, use a list of `backendAndTargets` in place of `backends` and `targets`.
- In wrappers, drop `setBackend` and `setTarget`, use `setBackendAndTarget` instead.
- In benchmark, use `setBackendAndTarget` in place of `setBackend` and `setTarget`.

fengyuentau:
Resolves #144


WanliZhong (approved):
👍👍👍We can use `cv.__version__ < '4.7.0'` directly to check OpenCV version.
image

fengyuentau:
Let me have a try on your solution.

fengyuentau:
👍

PR #147, opened 2023-03-17
Add person detection model from MediaPipe

WanliZhong:
This PR aims to add person detection model form MediaPipe. Pose estimation model will be added after this pr being merged.


WanliZhong:
cc @ShiqiYu, this is person detection model. Pose estimation still need few days to support half body pose estimation.

fengyuentau (approved):
👍

PR #157, opened 2023-05-04
beautify benchmark table

WanliZhong:
beautify the benchmark table with color map


benchmark/generate_table.py
```diff
@@ -0,0 +1,121 @@
...
+    # width of each column
+    col_widths = [0.25, 0.4, 0.15] + [0.15] * (len(hardware_info) - 3)
+    col_widths[7] = 0.2  # ascend310 is wider
+    # create a color bar. base width set to 1000 pixels, color map height set to 80 pixels
+    fig, axs = plt.subplots(nrows=2, figsize=(10, 0.8))
+    gradient = np.linspace(0, 1, 256)
+    gradient = np.vstack((gradient, gradient))
+    axs[0].imshow(gradient, aspect='auto', cmap=cmap)
+    axs[0].text(-0.01, 0.5, "BEST", va='center', ha='right', fontsize=11, transform=axs[0].transAxes)
+    axs[0].text(1.01, 0.5, "WORST", va='center', ha='left', fontsize=11, transform=axs[0].transAxes)
```
fengyuentau, line 100: `BEST` -> `Faster`
`WORST` -> `Slower`

WanliZhong: OK

benchmark/color_table.svg
fengyuentau, line 1: Just noticed that `faster` side is on the left and `slower` the opposite. Normally speaking, people read from left to right and they expect a `growth`, but the legend does the opposite. I suggest to put `slower` side on the left and `faster` side on the right. What do you think?

fengyuentau:
@ShiqiYu Could you take a look at the table if possible?

benchmark/color_table.svg
WanliZhong: I think it's true, but green is positive and red is negative. I think people prefer green than red.

benchmark/generate_table.py
```diff
@@ -0,0 +1,121 @@
+import re
+import matplotlib.pyplot as plt
+import matplotlib as mpl
```
fengyuentau, line 3: Please update `requirements.txt` to include `matplotlib` as well.

benchmark/README.md
```diff
@@ -57,6 +57,30 @@ python benchmark.py --all --cfg_overwrite_backend_target 1
...
+| [DaSiamRPN](../models/object_tracking_dasiamrpn)         | Object Tracking               | 1280x720   | 29.05                      | 712.94                     | 76.82                          | ---                     | ---                            | ---    |
+| [YoutuReID](../models/person_reid_youtureid)             | Person Re-Identification      | 128x256    | 30.39                      | 625.56                     | 90.07                          | 44.61                   | 5.58                           | ---    |
+| [MP-PalmDet](../models/palm_detection_mediapipe)         | Palm Detection                | 192x192    | 6.29                       | 86.83                      | 83.20                          | 33.81                   | 5.17                           | ---    |
+| [MP-HandPose](../models/handpose_estimation_mediapipe)   | Hand Pose Estimation          | 224x224    | 4.68                       | 43.57                      | 40.10                          | 19.47                   | 6.27                           | ---    |
+
+\*: Models are quantized in per-channel mode, which run slower than per-tensor quantized models on NPU.
```
fengyuentau, line 82: Add this note in the image as well?

benchmark/requirements.txt
```diff
@@ -1,4 +1,5 @@
 numpy
 opencv-python==4.5.4.58
```
fengyuentau, line 2: Should update to latest opencv-python. Maybe try `opencv-python<5.0`?

benchmark/color_table.svg
fengyuentau, line 1: I am thinking of a more meaningful header like this:

```
||               CPU                  ||       GPU       ||            NPU             ||
|| Intel 12700K | Rasberry Pi 4B | D1 ||  Jetson Nano B1 || Khadas VIM3 | Atals 200 DK ||
```

Just keep this in mind. This should be the next improvement.

benchmark/requirements.txt
WanliZhong: Yes, sure.

benchmark/color_table.svg
WanliZhong: I think it's a good idea. This can show which backend we should focus on.

fengyuentau (approved):
👍

PR #159, opened 2023-05-13
Fix broken files

fengyuentau:
Although this change looks like deleting files, actually they are deleted as git-tracked binary files. They are still available as git-lfs tracked files.

See the differences before and after this patch:
- Before: binary files, https://github.com/opencv/opencv_zoo/blob/master/models/text_recognition_crnn/text_recognition_CRNN_CH_2023feb_fp16.onnx
- After: LFS object, https://github.com/fengyuentau/opencv_zoo/blob/fix_broken_crnn_files/models/text_recognition_crnn/text_recognition_CRNN_CH_2023feb_fp16.onnx

**Need to think how to remove these binary files from history to reduce pull / clone size.**


WanliZhong (approved):
This problem also bothered me for a long time😊

PR #160, opened 2023-05-14
Add benchmark results on Toybrick RV1126, Khadas Edge2, Horizon Sunrise X3 PI and MAIX-III AX-PI

fengyuentau:
Todo:

- [x] add results on MAIX-III AX-PI
- [x] add npu desc for rv1126, khadas edge2 and maix-iii ax-pi
- [x] update benchmark table
- [x] also need to add benchmark results of the new models (MP-PersonDet)


WanliZhong:
Why we don't update the benchmark table?

fengyuentau:
> Why don't we update the benchmark table?

Two reasons:
1. These results are all from CPU, but these boards have NPUs that OpenCV does not support. These results should be provided as supplements.
2. The benchmark table should be upgraded to have CPU, GPU, NPU headers.

WanliZhong (approved):
👍LGTM

WanliZhong:
Merge this PR and need to update the table color to non-linear color map in time

PR #164, opened 2023-06-05
update generate table script

WanliZhong:
Update the color table:
- Automatically parse table data based on benchmark results
- Remove `(ms)` from the header
- Make color changes non-linear
- Header device direct to detail url
- Models name direct to detail url
- Split device companpy from device mdoel
- Make header text bold and center alignment
- Make color map be in the center of the table
- Add a config file for the parser:
  - acceptable_time: values greater than this will be set to red. I use the value of 100 times the best result
  - need other information for the parser

TODO: need @fengyuentau rename devices display_info


benchmark/table_config/table_config.yaml
fengyuentau, line 1: Just put this config in benchmark/. It does not worth creating a directory just for one file.

fengyuentau (approved):
👍👍

Issue #165, opened 2023-06-06
We are planning a force push recently and contributors should re-fork this repo


fengyuentau:
Two models are pushed as git-tracked binary files through https://github.com/opencv/opencv_zoo/pull/131 and this leads to a git history of an unreasonable size. So a force push is planned to remove these two files completely from git history, then add back as git-lfs-tracked files. Image files are also moving to git-lfs-tracked as well.

fengyuentau:
The force push was done just now. Removed files (example images and crnn fp16 models) will be added back as git-lfs tracked files later via pull requests. Also the default branch is renamed to `main`.

PR #166, opened 2023-06-08
Add back example output images as lfs tracked files

fengyuentau:
Part of https://github.com/opencv/opencv_zoo/issues/165


WanliZhong (commented):
The images in each model readme file are not ready since the path is change
image

fengyuentau:
> The images in each model readme file are not ready since the path is change image

Fixed.

WanliZhong (approved):
LGTM👍

PR #171, opened 2023-06-10
add mobilenet c++ sample

LaurentBerger:
I use this sample https://github.com/opencv/opencv/blob/4.x/samples/dnn/classification.cpp
For Cmakelist https://github.com/opencv/opencv/blob/4.x/samples/cpp/example_cmake/CMakeLists.txt


models/image_classification_mobilenet/labels_imagenet_1k.txt
fengyuentau, line 1: Can you put these labels in mobilenet.cpp? We made a decision that we do not store labels as a separate file.

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,140 @@
...
+        //! [Set input blob]
+        net.setInput(blob);
+        int classId;
+        double confidence;
+        Mat prob = net.forward();
+
+        //! [Get a class with a highest score]
+        Point classIdPoint;
+        minMaxLoc(prob.reshape(1, 1), 0, &confidence, 0, &classIdPoint);
+        classId = classIdPoint.x;
```
fengyuentau, line 132: Please do the following to improve readability:

```.diff
-       int classId;
-       double confidence;
        Mat prob = net.forward();

        //! [Get a class with a highest score]
        Point classIdPoint;
+       double confidence;
        minMaxLoc(prob.reshape(1, 1), 0, &confidence, 0, &classIdPoint);
-       classId = classIdPoint.x;
+       int classId = classIdPoint.x;
```

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,140 @@
...
+        std::string file = parser.get<String>("classes");
+        std::ifstream ifs(file.c_str());
+        if (!ifs.is_open())
+            CV_Error(Error::StsError, "File " + file + " not found");
+        std::string line;
+        while (std::getline(ifs, line))
+        {
+            classes.push_back(line);
+        }
+    }
```
fengyuentau, line 85: Please move the labels to the end of this file and replace this with a call to get labels.

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <opencv2/dnn.hpp>
+#include <opencv2/core/utils/logger.hpp>
+
+using namespace cv;
+using namespace dnn;
+
+std::string keys =
+"{ help  h          | | Print help message. }"
+"{ model m             | image_classification_mobilenetv1_2022apr.onnx | An optional path to file with preprocessing parameters }"
+"{ classes             | labels_imagenet_1k.txt |path to labels file  }"
```
fengyuentau, line 16: We do not need this if labels are moved to this file.

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,140 @@
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
```
fengyuentau, line 4: Not necessary.

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <opencv2/imgproc.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/dnn.hpp>
+#include <opencv2/core/utils/logger.hpp>
```
fengyuentau, line 8: It is not used right? I suggest using one header to replace all of these to keep it simple:

```
#include "opencv2/opencv.hpp"
```

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,140 @@
...
+"5: CUDA, "
+"6: WebNN }"
+"{ target           | 0 | Choose one of target computation devices: "
+"0: CPU target (by default), "
+"1: OpenCL, "
+"2: OpenCL fp16 (half-float precision), "
+"3: VPU, "
+"4: Vulkan, "
+"6: CUDA, "
+"7: CUDA fp16 (half-float preprocess) }";
```
fengyuentau, line 38: Can you align options with the Python demo?

https://github.com/opencv/opencv_zoo/blob/e6ad1212fb396ceba06f66fcf866536307a84247/models/image_classification_mobilenet/demo.py#L21-L34

fengyuentau (commented):
Thank you for contribution! Please take a look at my comments below.

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,1126 @@
...
+
+        //! [Create a 4D blob from a frame]
+        blobFromImage(frame, blob, scale, inpSize, mean, swapRB, crop);
+
+        // Check std values.
+        if (std.val[0] != 0.0 && std.val[1] != 0.0 && std.val[2] != 0.0)
+        {
+            // Divide blob by std.
+            divide(blob, std, blob);
+        }
```
fengyuentau, line 100: One suggestion:

In the upcoming OpenCV 4.8.0 release, there is a new API `blobFromImageWithParams` which applies mean and std when it loads from image. Here is the doc: https://docs.opencv.org/4.x/d6/d0f/group__dnn.html#ga8027ad777236a7ae543c6e4ffa4d544c.

Can you try with this API? You need to build the latest OpenCV though.

LaurentBerger: Yes but there is this [post ]([url]( https://forum.opencv.org/t/image2blobparams-c-and-python/13437))with associated [issue](https://github.com/opencv/opencv/issues/23764)

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,1126 @@
...
+
+    Net net = readNet(model);
+    net.setPreferableBackend(backendTargetPairs[backendTargetid].first);
+    net.setPreferableTarget(backendTargetPairs[backendTargetid].second);
+    //! [Open a video file or an image file or a camera stream]
+    VideoCapture cap;
+    if (parser.has("input"))
+        cap.open(parser.get<String>("input"));
+    else
+        cap.open(0, CAP_DSHOW);
```
fengyuentau, line 68: My webcam cannot be open with `CAP_DSHOW`. How about removing it and using just `cap.open(0)` instead?

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,1126 @@
+#include <opencv2/opencv.hpp>
```
fengyuentau, line 1: Also include vector, string and utility that you use in the following code:

```.diff
+ #include <vector>
+ #include <string>
+ #include <utility> // for std::pair and std::make_pair
#include <opencv2/opencv.hpp>

File: models/image_classification_mobilenet/mobilenet.cpp [hunk 10]
```diff
@@ -0,0 +1,1126 @@
...
+"{ initial_width    | 0 | Preprocess input image by initial resizing to a specific width.}"
+"{ initial_height   | 0 | Preprocess input image by initial resizing to a specific height.}"
+"{ rgb              | true | swap R and B plane.}"
+"{ crop             | false | Preprocess input image by center cropping.}"
+"{ backend bt         | 0 | Choose one of computation backends: "
+"0: (default) OpenCV implementation + CPU,"
+"1: CUDA + GPU (CUDA),"
+"2: CUDA + GPU (CUDA FP16),"
+"3: TIM-VX + NPU,"
+"4: CANN + NPU}";
```
fengyuentau, line 28: Could you align columns of these lines so that people can find options with default value and help message easily?

fengyuentau:
Could you also add a section in readme about how to build and use the c++ demo just like https://github.com/opencv/opencv_zoo/tree/main/models/face_detection_yunet#c does?

models/image_classification_mobilenet/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+cmake_minimum_required(VERSION 3.2)
+set(project_name "mobilenet")
```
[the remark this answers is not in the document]
LaurentBerger: Don't you want to change source file name (mobilenet.cpp to opencv_zoo_image_classification_mobilenet.cpp)?

models/image_classification_mobilenet/mobilenet.cpp
fengyuentau: Since we are making c++ demo, I think Python stuff is not a problem for us.

models/image_classification_mobilenet/mobilenet.cpp
```diff
@@ -0,0 +1,1126 @@
...
+    {
+        parser.printMessage();
+        return 0;
+    }
+
+    int rszWidth = parser.get<int>("initial_width");
+    int rszHeight = parser.get<int>("initial_height");
+    bool swapRB = parser.get<bool>("rgb");
+    bool crop = parser.get<bool>("crop");
+    String model = samples::findFile(parser.get<String>("model"));
```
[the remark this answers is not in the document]
fengyuentau: AFAIK, `samples::findfiles` is only used in `$opencv/samples`. We should avoid using it outside of its scope.

models/image_classification_mobilenet/CMakeLists.txt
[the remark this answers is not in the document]
fengyuentau: Just want to align with https://github.com/opencv/opencv_zoo/blob/e6ad1212fb396ceba06f66fcf866536307a84247/models/face_detection_yunet/CMakeLists.txt#L2. It is not a big problem though.

As for the source file name, I prefer a simpler name: `demo.cpp`.

models/image_classification_mobilenet/mobilenet.cpp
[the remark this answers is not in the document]
LaurentBerger: I'm not agree because 
demo --input=opencv-logo.png is easyer than demo --input=/opencv/samples/data/opencv-logo.png 
and you can call findFiles from python too
but I remove it 👍

models/image_classification_mobilenet/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+PROJECT (${project_name})
+
+find_package(OpenCV ${OPENCV_VERSION} REQUIRED HINTS ${OPENCV_INSTALLATION_PATH})
+# Find OpenCV, you may need to set OpenCV_DIR variable
+# to the absolute path to the directory containing OpenCVConfig.cmake file
+# via the command line or GUI
+
+set(OPENCV_VERSION "4.7.0")
+set(OPENCV_INSTALLATION_PATH "" CACHE PATH "Where to look for OpenCV installation")
+
```
fengyuentau, line 13: Line 11 and 12 should be placed before line 6:

```cmake
set(OPENCV_VERSION "4.7.0")
set(OPENCV_INSTALLATION_PATH "" CACHE PATH "Where to look for OpenCV installation")

find_package(OpenCV ${OPENCV_VERSION} REQUIRED HINTS ${OPENCV_INSTALLATION_PATH})
```

models/image_classification_mobilenet/demo.cpp
```diff
@@ -0,0 +1,1126 @@
...
+"4: CANN + NPU}";
+
+vector<string> getLabelImagenet1k();
+
+int main(int argc, char** argv)
+{
+    CommandLineParser parser(argc, argv, keys);
+
+    parser.about("Use this script to run classification deep learning networks in opencv Zoo  using OpenCV.");
+    if (argc == 1 || parser.has("help"))
```
fengyuentau, line 41: `argc == 1` can be removed since we can run with default values.

models/image_classification_mobilenet/demo.cpp
```diff
@@ -0,0 +1,1126 @@
...
+    }
+    vector<string> labels = getLabelImagenet1k();
+
+    Net net = readNet(model);
+    net.setPreferableBackend(backendTargetPairs[backendTargetid].first);
+    net.setPreferableTarget(backendTargetPairs[backendTargetid].second);
+    //! [Open a video file or an image file or a camera stream]
+    VideoCapture cap;
+    if (parser.has("input"))
+        cap.open(parser.get<String>("input"));
```
WanliZhong, line 66: `VideoCapture` can't open an image but used for reading a sequence of images with formatted name. I think we need to use `imread()` instead of.

models/image_classification_mobilenet/demo.cpp
```diff
@@ -0,0 +1,1126 @@
...
+            labels[classId].c_str()),
+            confidence);
+        putText(frame, label, Point(0, 55), FONT_HERSHEY_SIMPLEX, 0.5, Scalar(0, 255, 0));
+        imshow(kWinName, frame);
+    }
+    return 0;
+}
+
+
+vector<string> getLabelImagenet1k()
```
dkurt, line 120: Can the classes names be out of the scope of sample?

https://github.com/opencv/opencv/blob/e60a7c0d490b5bd94524680deb74177d7bfb9ae4/samples/data/dnn/classification_classes_ILSVRC2012.txt#L697

models/image_classification_mobilenet/demo.cpp
WanliZhong: I got this error, am I using it wrong?
```shell
❯ ./build/opencv_zoo_image_classification_mobilenet -i test.jpeg
[ERROR:0@0.019] global cap.cpp:166 open VIDEOIO(CV_IMAGES): raised OpenCV exception:

OpenCV(4.7.0-dev) /Users/wanli/Desktop/OpenCV_China/opencv/modules/videoio/src/cap_images.cpp:253: error: (-5:Bad argument) CAP_IMAGES: can't find starting number (in the name of file): true in function 'icvExtractPattern'


[ WARN:0@0.019] global cap.cpp:206 open VIDEOIO(CV_IMAGES): backend is generally available but can't be used to capture by name
libc++abi: terminating due to uncaught exception of type cv::Exception: OpenCV(4.7.0-dev) /Users/wanli/Desktop/OpenCV_China/opencv_zoo/models/image_classification_mobilenet/demo.cpp:70: error: (-2:Unspecified error) Cannot opend video or file in function 'main'

[1]    59294 abort      ./build/opencv_zoo_image_classification_mobilenet -i test.jpeg
```

models/image_classification_mobilenet/demo.cpp
fengyuentau: I think we just ship these labels along with the demo to make it work with less dependencies. For Python users, I would say very few of them have opencv source code. For C++ users, if they do not clone and build from source, they would not have opencv source code either.

models/image_classification_mobilenet/demo.cpp
LaurentBerger: No you can do :  
opencv_zoo_image_classification_mobilenet.exe -i=c:/lib/opencv/samples/data/lena.jpg -m=C:\lib\opencv_zoo\models\image_classification_mobilenet\image_classification_mobilenetv2_2022apr.onnx and result is 

image

models/image_classification_mobilenet/demo.cpp
LaurentBerger: Yes I can do it but in [previous review](https://github.com/opencv/opencv_zoo/pull/171#discussion_r1225326383) I put label name in source file.
Reviewer must choose

models/image_classification_mobilenet/mobilenet.cpp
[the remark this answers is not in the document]
fengyuentau: Frankly speaking, I thought `samples::findFile` was from `$opencv/samples` but it was defined and declared in the core module. Then I tried to search it in opencv source code, found that this API is only used in `$opencv/samples`.

Now my thought is these kinds of demos are meant to show beginers how to use models with opencv in the most simple way. In most of the cases, they would not even know `OPENCV_SAMPLES_DATA_PATH` but would find and input an absolute path of an image. I have to say it can be convenient for people who know it, but for other people it may cause confusion. We can use it if we want to promote it.

models/image_classification_mobilenet/demo.cpp
LaurentBerger: using c++ opencv parser I think we need an equal sign between option and arg : 
 ./build/opencv_zoo_image_classification_mobilenet -i=test.jpeg

models/image_classification_mobilenet/demo.cpp
dkurt: Thanks! This way, can we move them in a separate header and just `#include`?

models/image_classification_mobilenet/README.md
```diff
@@ -29,6 +31,24 @@ python demo.py --input /path/to/image --model v2
...
+
+```shell
+# A typical and default installation path of OpenCV is /usr/local
+cmake -B build -D OPENCV_INSTALLATION_PATH=/path/to/opencv/installation .
+cmake --build build
+
+# detect on camera input
+./build/opencv_zoo_image_classification_mobilenet
+# detect on an image
+./build/opencv_zoo_image_classification_mobilenet -m=/path/to/model -i=/path/to/image -v
```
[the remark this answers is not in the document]
LaurentBerger: Do you want to add an option to disable display like 
https://github.com/opencv/opencv_zoo/blob/main/models/text_detection_db/demo.py#L54-L55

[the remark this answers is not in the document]
LaurentBerger: Sorry I work on three git branch and I must remove rcnn files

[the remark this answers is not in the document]
WanliZhong: I think it is better to add it, because the headless version or server version of opencv cannot use GUI. Also keeps the c++ version and python version of the demo consistent.

models/image_classification_mobilenet/demo.cpp
LaurentBerger: @dkurt I have got a problem translating python to c++ code for [text_recognition_rcnn demo](https://github.com/opencv/opencv_zoo/blob/main/models/text_recognition_crnn/crnn.py#L85)
In python code net.forward output is a numpy array : shape is known using outblob.shape

in C++ how can i guess blob size?

[here I put 24](https://github.com/LaurentBerger/opencv_zoo/blob/text_recognition_crnn/models/text_recognition_crnn/demo.cpp#L160) because I ran python code. It is for model text_recognition_CRNN_EN*.onnx
for model text_recognition_CRNN_CN_2021 and text_recognition_CRNN_CH I must use 25. Is it possible to guess those values without check model name?

dkurt: @LaurentBerger, you may access all dimensions by `size` array: `outputBlob.size[i]`. Opositting to `outputBlob.size()` it's multidimensional.

WanliZhong (approved):
👍LGTM!

fengyuentau (approved):

Issue #181, opened 2023-06-21
handpose 


hlym123:
运行手势识别demo时，手在窗口靠右侧时无法识别。应该是  pose_estimation_mediapipe/mp_handpose.py 代码有误

```python
# line 57
palm_bbox[:, 0] = np.clip(palm_bbox[:, 0], 0, image.shape[0])  # image.shape -- h,w,c 
palm_bbox[:, 1] = np.clip(palm_bbox[:, 1], 0, image.shape[1])
```
改为
```python
palm_bbox[:, 0] = np.clip(palm_bbox[:, 0], 0, image.shape[1])  
palm_bbox[:, 1] = np.clip(palm_bbox[:, 1], 0, image.shape[0])
```

fengyuentau:
cc @WanliZhong

WanliZhong:
@hlym123 Thanks for your report, I think it's really a mistake I have made. I will fix it soon.

PR #182, opened 2023-06-23
HandPose Estimation: clip palm bbox to (width, height)

WanliZhong:
Fixes bug #181 

palm box is `((x1, y1), (x2, y2))`, so it should be clipped to `(image.shape[1], image.shape[0])`


fengyuentau (approved):
👍

PR #184, opened 2023-06-24
update accuracy evaluation scripts

WanliZhong:
Some information about models or evaluation scripts are outdated. This PR aims at updating them. Merged after #185


tools/eval/datasets/imagenet.py
```diff
@@ -63,3 +64,1004 @@ def get_result(self):
     def print_result(self):
         print("Top-1 Accuracy: {:.2f}%; Top-5 Accuracy: {:.2f}%".format(self.top1_acc*100, self.top5_acc*100))
 
+
+    LABELS_IMAGENET_1K = '''tench
```
fengyuentau, line 68: We dont need these labels for evaluation.

WanliZhong: But the `model.infer(img)` will give labels directly not index of labels.

fengyuentau: I suggest to add `load_label=True` in for example `PPResNet` then set it `False` in evaluation.

WanliZhong: I will create another PR to modified it.

WanliZhong: What's the difference between `text_recognition_CRNN_EN_2021sep.onnx` and `text_recognition_CRNN_EN_2023feb.onnx`? They have the same accuracy through evaluation.

fengyuentau (approved):
LGTM 👍

PR #217, opened 2023-11-14
YOLOX graph with decoding for test

Abdurrahheem:
This PR adds onnx graph of yolox with decoding. The graph will be used in OpenCV onnx imported tests. Onnx graphs that are in model zoo do not contain decoding in them. Decoding inside onnx graph is required to reduce additional functionally of creating anchors inside tests

### DIFF
--- a/models/object_detection_yolox/object_detection_yolox_decoder.onnx
+++ b/models/object_detection_yolox/object_detection_yolox_decoder.onnx
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:2c893e3d08c3c71dc0bd09b32c6677bf9acd65d69e4fcb0cef5a97e13b314aa1
+size 36014058


dkurt:
@Abdurrahheem, thanks! Are we sure that this model already supported by OpenCV or can be supported in future?

models/object_detection_yolox/object_detection_yolox_decoder.onnx
dkurt, line 1: What is the source of the model? Does it make sense to add conversion step to README?

Abdurrahheem:
> @Abdurrahheem, thanks! Are we sure that this model already supported by OpenCV or can be supported in future?

Yes, it is supported

models/object_detection_yolox/object_detection_yolox_decoder.onnx
Abdurrahheem: Probably yes

fengyuentau: Also what do you mean by the `_decoder` suffix? What is the main differences between your model and the existing one?

ShiqiYu:
OpenCV zoo is for mature models, which should be small and efficient, and easy to be deploy by users.

PR #225, opened 2023-12-28
add benchmark results on Atlas 200I DK A2

fengyuentau:
Currently we have problem with NPU drivers (not opencv issues). After it is figured out we will add results on the NPU Ascend 310B.


WanliZhong (approved):
👍

PR #226, opened 2023-12-28
add benchmark for optical flow Raft model

WanliZhong:


WanliZhong:
Fail on CANN backend. 
```
[ WARN:0@5.268] global net_cann.cpp:120 initBackend DNN/CANN: layer (name=onnx_node!InstanceNormalization_16/MVN, type=MVN) is not supported by CANN backend. Going back to default backend on CPU target
```

fengyuentau:
> Fail on CANN backend.
> 
> ```
> [ WARN:0@5.268] global net_cann.cpp:120 initBackend DNN/CANN: layer (name=onnx_node!InstanceNormalization_16/MVN, type=MVN) is not supported by CANN backend. Going back to default backend on CPU target
> ```

It should not go to MVN. We have dedicated instance norm layer already.

WanliZhong:
Benchmark is already done. Raft run very very slow. I set 30s as the accepted time for the color table.

fengyuentau:
Thanks for the hard work. I propose not to put this model into the benchmark table. Lets discuss in the meeting.

Issue #135, opened 2023-02-22
Add C++ demos for models in the zoo


fengyuentau:
We have provided at least one easy and clean Python demo for every model here in the zoo. As described in https://github.com/opencv/opencv_zoo/issues/132, Python demos sometimes can be too concise to be converted to other languages, such as C++. Hence, we decide to add C++ demos as well but they should be **clean and simple enough to show how to run inference and get expected output with OpenCV**.

**We welcome contributions from community**. Please take a look at the list below and leave comments for application or discussion before you start to dive in coding.

| Status | Task | Models |
| - | - | - |
| 🏃 In-progress | Face Detection | YuNet |
| ❗️ Need Contribution | Face Recognition | SFace |
| ❗️ Need Contribution | Object Detection | YOLOX |
| ❗️ Need Contribution | Object Detection | NanoDet |
| ❗️ Need Contribution | Text Detection | DB |
| ❗️ Need Contribution | Text Recognition | CRNN (CN) |
| ❗️ Need Contribution | Text Recognition | CRNN (EN) |
| ❗️ Need Contribution | Text Recognition | CRNN (CH) |
| ❗️ Need Contribution | Image Classification | PP-ResNet50 |
| ❗️ Need Contribution | Image Classification | MobileNet V1 |
| ❗️ Need Contribution | Image Classification | MobileNet V2 |
| ❗️ Need Contribution | Human Segmentation | PP-HumanSeg |
| ❗️ Need Contribution | QR Code Detection / Parsing | WeChatQRCode |
| ❗️ Need Contribution | Object Tracking | DaSiamRPN |
| ❗️ Need Contribution | Person Re-Identification | YoutuReID |
| ❗️ Need Contribution | Palm Detection | MP-PalmDet |
| ❗️ Need Contribution | Hand Pose Estimation | MP-HandPose |
| ❗️ Need Contribution | Facial Expression Recognition | FER |

kekxv:
[handpose_estimation_mediapipe.](https://github.com/opencv/opencv_zoo/blob/master/models/handpose_estimation_mediapipe) please
Thank you so much.

fengyuentau:
It would be good to provide additional demos in G-API to construct a complete end-to-end pipeline. But first we need to have DNN backend in G-API, which is not implemented yet right?

cc @ShiqiYu

LaurentBerger:
[...]

PR #234, opened 2024-02-26
Fix OpenCV version assertions in ViTTrack & RAFT and text offset

ryan1288:
This PR addresses compatibility issues with OpenCV version 4.9.0 by updating assertions in the code. Issue was pointed out in https://github.com/opencv/opencv_zoo/issues/135#issuecomment-1962323969. 

Additionally, a minor adjustment has been made to increase the offset, preventing the confidence text from overlapping with the bounding box border.
image


fengyuentau (approved):
Thank you for the contribution 👍

PR #235, opened 2024-02-26
Add YOLOX C++ demo save functionality

ryan1288:
Minor PR to add missing save functionality to the YOLOX C++ demo. Issue was pointed out in https://github.com/opencv/opencv_zoo/issues/135#issuecomment-1962323969.

### DIFF
--- a/models/object_detection_yolox/demo.cpp
+++ b/models/object_detection_yolox/demo.cpp
@@ -267,6 +267,7 @@ int main(int argc, char** argv)
     bool vis = parser.get<bool>("vis");
     bool save = parser.get<bool>("save");
     int backendTargetid = parser.get<int>("backend");
+    bool isInputImageProvided = parser.has("input");
 
     if (model.empty())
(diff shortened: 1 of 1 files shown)


models/object_detection_yolox/demo.cpp
```diff
@@ -306,6 +306,10 @@ int main(int argc, char** argv)
         tm.stop();
         cout << "Inference time: " << tm.getTimeMilli() << " ms\n";
         Mat img = visualize(predictions, frame, letterboxScale, tm.getFPS());
+        if (save)
```
fengyuentau, line 309: It is not good saving as `.jpg` when the input is camera stream. It just keeps overwriting the last one.

ryan1288: Agreed. I've updated it to only save an image if the user has an input image.

fengyuentau (approved):
LGTM 👍

PR #236, opened 2024-02-26
Fix spelling, spacing, and unused variables

ryan1288:
Minor PR fixing spelling mistakes, spacing, and removed unused variables. Issue partially mentioned in https://github.com/opencv/opencv_zoo/issues/135#issuecomment-1962323969.

### DIFF
--- a/models/person_detection_mediapipe/demo.cpp
+++ b/models/person_detection_mediapipe/demo.cpp
@@ -237,10 +237,9 @@ int main(int argc, char** argv)
         backendTargetPairs[backendTargetid].first, backendTargetPairs[backendTargetid].second);
     //! [Open a video file or an image file or a camera stream]
     if (!cap.isOpened())
-        CV_Error(Error::StsError, "Cannot opend video or file");
+        CV_Error(Error::StsError, "Cannot  video or file");
 
     static const std::string kWinName = "MPPersonDet Demo";
-    int nbInference = 0;
     while (waitKey(1) < 0)
     {
         cap >> frame;
--- a/models/person_detection_mediapipe/demo.cpp
+++ b/models/person_detection_mediapipe/demo.cpp
@@ -43,14 +43,6 @@ public:
         this->anchors = getMediapipeAnchor();
     }
 
-    void setBackendAndTarget(dnn::Backend bId, dnn::Target tId)
-    {
-        this->backendId = bId;
-        this->targetId = tId;
-        this->net.setPreferableBackend(this->backendId);
-        this->net.setPreferableTarget(this->targetId);
-    }
-
     pair<Mat, Size> preprocess(Mat img)
     {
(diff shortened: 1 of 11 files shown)


ryan1288:
Additional note: While I did not implement this change, I observed that the following was declared but unused in the C++ demos.
```
void setBackendAndTarget(dnn::Backend bId, dnn::Target tId)
{
    this->backendId = bId;
    this->targetId = tId;
    this->net.setPreferableBackend(this->backendId);
    this->net.setPreferableTarget(this->targetId);
}
```

models/person_detection_mediapipe/demo.cpp
```diff
@@ -237,10 +237,9 @@ int main(int argc, char** argv)
         backendTargetPairs[backendTargetid].first, backendTargetPairs[backendTargetid].second);
     //! [Open a video file or an image file or a camera stream]
     if (!cap.isOpened())
-        CV_Error(Error::StsError, "Cannot opend video or file");
+        CV_Error(Error::StsError, "Cannot  video or file");
```
fengyuentau, line 240: open

fengyuentau:
> Additional note: While I did not implement this change, I observed that the following was declared but unused in the C++ demos.
> 
> ```
> void setBackendAndTarget(dnn::Backend bId, dnn::Target tId)
> {
>     this->backendId = bId;
>     this->targetId = tId;
>     this->net.setPreferableBackend(this->backendId);
>     this->net.setPreferableTarget(this->targetId);
> }
> ```

Which particular demo?

models/person_detection_mediapipe/demo.cpp
ryan1288: Fixed! Also updated several more typos.

ryan1288:
Models with that unused function:

- `face_detection_yunet/demo.cpp`
- `object_detection_yolox/demo.cpp`
- `person_detection_mediapipe/demo.cpp`
- 2x in `pose_estimation_mediapipe/demo.cpp` for each class

In the Python equivalent classes and `benchmark.py`, it is used for benchmarking. However, currently, they are unused in the C++ versions. Do we plan to keep them for future evaluation?

fengyuentau:
> In the Python equivalent classes and benchmark.py, it is used for benchmarking. However, currently, they are unused in the C++ versions. Do we plan to keep them for future evaluation?

Well, you are right. They are used for benchmark in Python scripts but we don't need them in C++ demos. It's okay to remove them then.

ryan1288:
@fengyuentau  I've removed all such instances of unused `setBackendAndTarget` from C++ demos. I also reran all the C++ demos to ensure functionality was not affected.

fengyuentau (approved):

fengyuentau:
Thank you for your contritbution 👍

PR #237, opened 2024-02-26
Fix visualization and default model for text detection PP-OCR

ryan1288:
Minor PR to correct the default model path and address issues with visualization. The issue was partially discussed in https://github.com/opencv/opencv_zoo/issues/135#issuecomment-1962323969.

### DIFF
--- a/models/text_detection_ppocr/demo.cpp
+++ b/models/text_detection_ppocr/demo.cpp
@@ -18,7 +18,7 @@ vector< pair<cv::dnn::Backend, cv::dnn::Target> > backendTargetPairs = {
 
 std::string keys =
 "{ help  h           |                                              | Print help message. }"
-"{ model m           | text_detection_ch_ppocrv3_2023may.onnx       | Usage: Set model type, defaults to text_detection_ch_ppocrv3_2023may.onnx }"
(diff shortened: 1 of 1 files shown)


fengyuentau:
@WanliZhong Please take a look at this PR.

models/text_detection_ppocr/demo.cpp
```diff
@@ -172,7 +172,10 @@ int main(int argc, char** argv)
                 imwrite("result.jpg", originalImage);
             }
             if (viz)
+            {
                 imshow(kWinName, originalImage);
+                waitKey(0);
```
WanliZhong, line 177: Line 30 has
```c++
 while (waitKey(1) < 0)
```
So I think we don‘t need to add `waitKey(0)`?

ryan1288: `while (waitKey(1) < 0)` - I believe this has a timeout of 1 ms before continuing the loop.

It then attempts to go into 
```
cap >> originalImage;
if (originalImage.empty())
{
    if (parser.has("input"))
    {
        cout << "Frame is empty" << endl;
        break;
    }
    else
        continue;
}
```
and reaches `cout << "Frame is empty" << endl;`. On the otherhand, waitKey(0) waits indefinitely, allowing the user to keep the processed image up for viewing.

WanliZhong (approved):
LGTM👍

PR #240, opened 2024-03-03
C++ Demo - Object Tracking (VitTrack)

ryan1288:
This PR adds the C++ demo for the Object Tracking VitTrack model, as part of a request from https://github.com/opencv/opencv_zoo/issues/135.

- `demo.cpp`: Added to have matching functionality to the `demo.py`, text offsets in both versions were adjusted to prevent overlap and improve clarity.
- `demo.py`: Updated to use save and vis flags for optional saving (disabled by default) and visualization (enabled by default) functionality.
- `CMakeLists.txt`: Created to build with OpenCV 4.9.0 and enforces C++11 standards upon building.
- `README.md`: Updated with C++ instructions.

The interface matches other C++ demos within the set of ML models.

Average 100+ FPS on my laptop (AMD Ryzen 7 5800H) for a (640 x 480) webcam feed. FPS is slightly higher on a (1273 x 733) input video, see testing example below.

Testing

- [x] Run demo using webcam
```
cmake -B build -D OPENCV_INSTALLATION_PATH=/path/to/opencv .
cmake --build build
./build/opencv_zoo_object_tracking_vittrack
```
phone

- [x] Run demo using input video
```
cmake -B build -D OPENCV_INSTALLATION_PATH=/path/to/opencv .
cmake --build build
./build/opencv_zoo_object_tracking_vittrack -i=dog.mp4
```
dog

### DIFF
--- a/models/object_tracking_vittrack/demo.cpp
+++ b/models/object_tracking_vittrack/demo.cpp
@@ -12,7 +12,8 @@ struct TrackingResult
     float score;
 };
 
-class VitTrack {
+class VitTrack
+{
 public:
 
     VitTrack(const string& model_path, int backend_id = 0, int target_id = 0) 
@@ -92,7 +93,8 @@ int main(int argc, char** argv)
     bool save = parser.get<bool>("save");
     bool vis = parser.get<bool>("vis");
 
-    vector<vector<int>> backend_target_pairs = {
+    vector<vector<int>> backend_target_pairs =
+    {
         {DNN_BACKEND_OPENCV, DNN_TARGET_CPU},
         {DNN_BACKEND_CUDA, DNN_TARGET_CUDA},
         {DNN_BACKEND_CUDA, DNN_TARGET_CUDA_FP16},
--- a/models/object_tracking_vittrack/demo.cpp
+++ b/models/object_tracking_vittrack/demo.cpp
@@ -1,27 +1,30 @@
 #include <iostream>
 #include <opencv2/opencv.hpp>
 
+using namespace std;
+using namespace cv;
+
 struct TrackingResult {
     bool isLocated;
-    cv::Rect bbox;
+    Rect bbox;
     float score;
 };
 
 class VitTrack {
 public:
 
-    VitTrack(const std::string& model_path, int backend_id = 0, int target_id = 0) {
+    VitTrack(const string& model_path, int backend_id = 0, int target_id = 0) {
         params.net = model_path;
         params.backend = backend_id;
         params.target = target_id;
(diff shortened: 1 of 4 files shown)


ryan1288:
@fengyuentau I noticed that the "save" functionality should be optionally enabled to save an output `.mp4` with the tracking bounding box. This is currently not available in neither the Python demo (nor the new C++ demo yet). I'll add that functionality to both C++ and Python in a different PR. Other models may have the same problem so I'll add it to them as well.

models/object_tracking_vittrack/demo.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+int main(int argc, char** argv) {
+    CommandLineParser parser(argc, argv,
+        "{help  h           |                                       | Print help message. }"
+        "{input i           |                                       |Set path to the input video. Omit for using default camera.}"
+        "{model_path        |object_tracking_vittrack_2023sep.onnx  |Set model path}"
+        "{backend_target bt |0                                      |Choose backend-target pair: 0 - OpenCV implementation + CPU, 1 - CUDA + GPU (CUDA), 2 - CUDA + GPU (CUDA FP16), 3 - TIM-VX + NPU, 4 - CANN + NPU}"
+        "{save s            |false                                  |Specify to save a file with results. Invalid in case of camera input.}"
+        "{vis v             |false                                  |Specify to open a new window to show results. Invalid in case of camera input.}");
```
fengyuentau, line 71: Seems like both of them are not used anywhere.

fengyuentau:
> @fengyuentau I noticed that the "save" functionality should be optionally enabled to save an output `.mp4` with the tracking bounding box. This is currently neither available in the Python demo nor the new C++ demo. I'll add that functionality to both C++ and Python in a different PR. Other models may have the same problem so I'll add it to them as well.

Ok, it would be like to have this fuctionality implemented.

models/object_tracking_vittrack/demo.cpp
ryan1288: They are now used! :+1: In both C++ and Python.

ryan1288:
> > @fengyuentau I noticed that the "save" functionality should be optionally enabled to save an output `.mp4` with the tracking bounding box. This is currently neither available in the Python demo nor the new C++ demo. I'll add that functionality to both C++ and Python in a different PR. Other models may have the same problem so I'll add it to them as well.
> 
> Ok, it would be like to have this fuctionality implemented.

The functionality is added :100: Saving works for both input videos and camera feed with matching FPS.

fengyuentau (approved):
👍 LGTM

PR #241, opened 2024-03-04
C++ Demo - Image Classification (PPResNet)

ryan1288:
This PR adds the C++ demo for the Image Classification PPResNet model, as part of a request from https://github.com/opencv/opencv_zoo/issues/135.

- `demo.cpp`: Added to have matching functionality to `demo.py`. All 1,000 classes included as a string to be parsed.
- `demo.py`: Printout was adjusted be cleaner and matches the C++ version. 
- `CMakeLists.txt`: Created to build with OpenCV 4.9.0 and enforces C++11 standards upon building.
- `README.md`: Updated with C++ instructions.

The interface matches other C++ demos within the set of ML models. 

The predictions visualizations were adjusted for both C++ and Python demos for clarity.
Before (Python)
image
After (Python & C++)
image
image

Testing
- [x] Matching Python and C++ demo classifcations
A number of images were tested with top-k = 20 and all had matching output classes, with one example provided below.
goldfish
image
- [x] Run demo with top-k = 1
```
cmake -B build -D OPENCV_INSTALLATION_PATH=/path/to/opencv .
cmake --build build
./build/opencv_zoo_image_classification_ppresnet --input=example_input/goldfish.jpg
```
image
- [x] Run demo with top-k > 1
```
cmake -B build -D OPENCV_INSTALLATION_PATH=/path/to/opencv .
cmake --build build
./build/opencv_zoo_image_classification_ppresnet --input=example_input/goldfish.jpg --top_k=5
```
image


models/image_classification_ppresnet/labels.txt
fengyuentau, line 1: Merge this file with `demo.cpp` and revert the changes in `ppresnet.py`. We made a decision long time ago that demo should be portable and convenient as much as possible with less files.

models/image_classification_ppresnet/example_input/goldfish.jpg
fengyuentau, line 1: I don't think we need example input as it demonstrates nothing and it increase repo size.

models/image_classification_ppresnet/labels.txt
ryan1288: Updated! I've reverted the changes to `ppresnet.py` and merged the labels with `demo.cpp`. Agree with the portability  :smile: .

models/image_classification_ppresnet/example_input/goldfish.jpg
ryan1288: Removed example :+1:

models/image_classification_ppresnet/demo.cpp
```diff
@@ -0,0 +1,1132 @@
...
+        for (size_t i = 0; i < predictions.size(); ++i) 
+        {
+            cout << "(" << i+1 << ") " << predictions[i] << endl;
+        }
+    }
+
+    return 0;
+}
+
+string LABELS_IMAGENET_1K = "tench\n"
```
fengyuentau, line 133: How about using `std::vector<std::string>`? In this case we do not need to duplicate `\n` for each class and we do not need to re-load this in `loadLabels`.

ryan1288: @fengyuentau  That's a great idea. Not sure why I didn't think of that :sweat_smile: I've updated it.

fengyuentau (approved):
Thank you for the contribution! 👍

PR #246, opened 2024-03-23
Update demo.py

ChiragBindal:


ChiragBindal:
Added threads so that it can work asyncronously

models/image_classification_mobilenet/demo.py
```diff
@@ -50,6 +50,15 @@
 
     # Inference
     result = model.infer(image)
+    return result
+
+if __name__ == '__main__':
+    # multiple image paths as input, processes them asynchronously using a thread pool, and prints the results . Caution only if large datasets
+    with concurrent.futures.ThreadPoolExecutor() as executor:
```
fengyuentau, line 57: Demo is meant to be easy. We don't use this in a demo.

PR #258, opened 2024-05-16
Migrating the EfficientSAM model to the OpenCV model zoo

Zhang-Yang-Sustech:
# EfficientSAM Model Demonstration

I have successfully written the code for the EfficientSAM model and created an interactive demo that enables normal model inference and user interaction. Users can easily use this demo via the command line with the following command:

```bash
python demo.py --input /path/to/your/image.jpg
```

By simply providing the path to the image they wish to segment, the demo will display the image. Upon clicking the object they want to segment in the image, a new window will show the segmentation result, making it user-friendly and straightforward.

## Example

example1

Please note that the current model inference speed takes approximately 2 seconds. The camera demo is not yet completed, and I plan to optimize it and add camera functionality in the future.

## Performance Testing

The model's benchmark testing has not been conducted yet. I will perform the tests in the future and provide the relevant performance data.

---

Thank you for your attention. I will continue to update and optimize the EfficientSAM model and its demonstration.

PR #260, opened 2024-06-03
bump opencv version to 4.10.0

fengyuentau:


fengyuentau:
Version control is broken when the minor version becomes two digit.

```
$ python
>>> "4.10.0" >= "4.9.0"
>>> False
```

It affects users who are using latest OpenCV 4.x (4.10.0, for example) but opencv_zoo 4.9.

fengyuentau:
This version brings bug fixes and some performance optimizations specifically for RISC-V with RVV extention, which is not existing in the hardware zoo. So no updates on the benchmark stats. Will Merge as-is.

WanliZhong (commented):
Missed modification of the `EfficientSAM` from https://github.com/opencv/opencv_zoo/pull/258

WanliZhong (approved):

Issue #251, opened 2024-04-07
Add Java demos (Last update on 2024-04-07)


fengyuentau:
There are a growing number of issues asking reproducing demos in Java. We welcome your contributions!

Relates https://github.com/opencv/opencv_zoo/issues/135.

Progress:

| Status | Task | Models |
| - | - | - |
| ❗️ Need Contribution | Face Detection | YuNet |
| ❗️ Need Contribution | Face Recognition | SFace |
| ❗️ Need Contribution | Object Detection | YOLOX |
| ❗️ Need Contribution | Object Detection | NanoDet |
| ❗️ Need Contribution | Text Detection | DB |
| ❗️ Need Contribution | Text Recognition | CRNN (CN) |
| ❗️ Need Contribution | Text Recognition | CRNN (EN) |
| ❗️ Need Contribution | Text Recognition | CRNN (CH) |
| ❗️ Need Contribution | Image Classification | PP-ResNet50 |
| ❗️ Need Contribution | Image Classification | MobileNet V1 |
| ❗️ Need Contribution | Image Classification | MobileNet V2 |
| ❗️ Need Contribution | Human Segmentation | PP-HumanSeg |
| ❗️ Need Contribution | QR Code Detection / Parsing | WeChatQRCode |
| ❗️ Need Contribution | Person Re-Identification | YoutuReID |
| ❗️ Need Contribution | Palm Detection | MP-PalmDet |
| ❗️ Need Contribution | Hand Pose Estimation | MP-HandPose |
| ❗️ Need Contribution | Person Detection | MP-PersonDet |
| ❗️ Need Contribution | Pose Estimation | MP-Pose |
| ❗️ Need Contribution | Facial Expression Recognition | FER |
| ❗️ Need Contribution | Object Tracking | VitTrack |

fengyuentau:
SFace Java demo: https://github.com/opencv/opencv_zoo/issues/195.

12341234-yangboran:
How can i upload my demo

12341234-yangboran:
Can I push the demo to branch of master?

fengyuentau:
> Can I push the demo to branch of master?

Yes, please make a pull request to add your java demo.

Oscar125841:
I have my demo.java of face_detection_yunet may I do the pull request

PR #264, opened 2024-06-28
Add Java demo-Text Detection, Object Tracking

bluehatch:
This PR is about creating Java demos (#251). The implementation model is as follows:
- Text Detection
- Object Tracking

The Java demo used OpenCV's java implementation [JavaCV](https://github.com/bytedeco/javacv).

I will continue developing Java demos for other models in the future.

### DIFF
--- a/models/object_tracking_vittrack/demo.java
+++ b/models/object_tracking_vittrack/demo.java
@@ -1,0 +1,206 @@
+import com.beust.jcommander.JCommander;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.UnixStyleUsageFormatter;
+import org.bytedeco.javacpp.BytePointer;
+import org.bytedeco.opencv.global.opencv_dnn;
+import org.bytedeco.opencv.opencv_core.*;
+import org.bytedeco.opencv.opencv_video.TrackerVit;
+import org.bytedeco.opencv.opencv_videoio.VideoCapture;
+import org.bytedeco.opencv.opencv_videoio.VideoWriter;
+
+import static org.bytedeco.opencv.global.opencv_highgui.*;
+import static org.bytedeco.opencv.global.opencv_imgproc.*;
+import static org.bytedeco.opencv.global.opencv_videoio.CAP_PROP_FPS;
+
+public class demo {
+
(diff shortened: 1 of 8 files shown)


fengyuentau:
I never use Java with Maven before. Could you explain what thoese `pom.xml` files are? Are they needed by Maven? How are they generated if needed?

fengyuentau (commented):
Thank you for contribution 👍

---

I never use Java with Maven before. Could you explain what thoese pom.xml files are? Are they needed by Maven? How are they generated if needed?

---

Also I propose to submit a demo for a model in a PR. So could you split this one into two?

bluehatch:
> Thank you for contribution 👍
> 
> I never use Java with Maven before. Could you explain what thoese pom.xml files are? Are they needed by Maven? How are they generated if needed?
> 
> Also I propose to submit a demo for a model in a PR. So could you split this one into two?

pom.xml is a file that contains Maven build information. It is usually written directly by the developer.
You can install Maven [here](https://maven.apache.org/download.cgi). For convenience, let's consider including a maven wrapper that can be used without installing Maven.

As requested, I will split the PR into two models and include a demo for each.

models/object_tracking_vittrack/demo.java
```diff
@@ -0,0 +1,206 @@
+import com.beust.jcommander.JCommander;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.UnixStyleUsageFormatter;
+import org.bytedeco.javacpp.BytePointer;
+import org.bytedeco.opencv.global.opencv_dnn;
+import org.bytedeco.opencv.opencv_core.*;
```
opencv-alalek, line 6: > org.bytedeco.opencv

This doesn't look as original OpenCV Java API.

/cc @asmorkalov

asmorkalov: Yes, it's 3rdparty package. We do not distribute pre-built Java package with Maven Central.

fengyuentau: > We do not distribute pre-built Java package with Maven Central

So there is nothing for Java like Python does (pip install then use out-of-the-box)?

PR #265, opened 2024-07-07
[GSoC] Blockwise Quantization Tool

DaniAffCH:
This PR introduces a Python tool for block quantizing ONNX models. The quantized models adhere to the ONNX standard, verified using `onnx.checker.check_model(self.model, full_check=True)`.

Additionally, these block-quantized models are compatible with the `QuantizeLinear` and `DequantizeLinear` layers in OpenCV, as introduced in opencv/opencv#25644, allowing them to be executed within the OpenCV DNN engine.

The tool currently performs asymmetric weight-only quantization on Convolutional Layers. It's possible to specify the desired quantization block size.
The quantization is applied along axis 1, flattening convolutional weights $[C_{out}, C_{in}, K_w, K_h] \rightarrow [C_{out}, C_{in} \times K_w \times K_h]$. 
Future enhancements could extend the tool capabilities, making it more customizable and general.

The tool also provides a quantization summary, reporting the overall quantization mean squared error and the initial and final model size. 

image

## Testing

When employing a block size of 16 and normalized input images, the mean squared quantization error was found to be in the order of magnitude of $10^{-2}$ or $10^{-3}$.

Furthermore, a qualitative assessment was conducted by quantizing some models of this repository in a blockwise manner and executing them, comparing the results with the original model and int8 model.

The findings indicate that, with even block size, the block quantized model maintains performance levels equivalent to those of the original model while achieving a reduction in model size.

Here is an example applied to [face detection yunet](https://github.com/opencv/opencv_zoo/tree/main/models/face_detection_yunet):
_Loading the following gif may take some time because of the gif size_ 

anim-opt

## Onnxruntime and DNN comparison

The resulting models have been tested using both `onnxruntime` and `OpenCV DNN`, both of which produced identical outputs for the same input data. 
Since onnxruntime introduced the support for blockwise quantization inference recently and such functionality has not been included in the last release, the only way to test it is to build onnxruntime from the source. 

```bash
./build.sh --config=Release --build_shared_lib --disable_ml_ops --build_wheel --enable_reduced_operator_type_support --skip_tests --parallel
```

Then install with pip the resulting wheel:

```bash
pip install -U build/Linux/Release/dist/onnxruntime-1.19.0-cp39-cp39-linux_x86_64.whl
```

---

To test the resulting networks with OpenCV DNN you have to build the pull request opencv/opencv#25644

### DIFF
--- a/tools/quantize/block_quantize.py
+++ b/tools/quantize/block_quantize.py
@@ -16,13 +16,10 @@ from onnx import helper
 
 BITS_TO_NUMPY_TYPE = {8: np.uint8, 16: np.uint16}
 
-CONVOLUTION = "Conv"
 
-KERNEL_SHAPE = "kernel_shape"
-STRIDE = "strides"
-PADS = "pads"
-DILATIONS = "dilations"
-GROUP = "group"
+SUPPORTED_OPS = {
+    "Conv"
+}
 
 ONNX_OPSET = 21
 
@@ -47,12 +44,7 @@ class BlockQuantizeResult:
 
 
 @dataclass
-class ConvParams:
-    kernel_shape: List[int] = field(default_factory=list)
-    strides: List[int] = field(default_factory=list)
-    pads: List[int] = field(default_factory=list)
-    dilations: List[int] = field(default_factory=list)
-    group: int = 1
+class LayerParams:
     weights: np.ndarray = field(default_factory=lambda: np.array([]))
     bias: Optional[np.ndarray] = None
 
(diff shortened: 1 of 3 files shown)


tools/quantize/block_quantize.py
```diff
@@ -0,0 +1,432 @@
...
+
+BITS_TO_NUMPY_TYPE = {8: np.uint8, 16: np.uint16}
+
+CONVOLUTION = "Conv"
+
+KERNEL_SHAPE = "kernel_shape"
+STRIDE = "strides"
+PADS = "pads"
+DILATIONS = "dilations"
+GROUP = "group"
```
fengyuentau, line 18: Could you pack them in a dictionary so that it can be more generic. Related code should be changed accordingly for generalization so that we can easily apply the similar quantization to `MatMul` or `Gemm`.

tools/quantize/block_quantize.py
```diff
@@ -0,0 +1,432 @@
...
+                f"Bits must be one of the following values: [{allowed_values}]."
+            )
+
+    def get_initializer_tensor(self, name: str) -> Optional[np.ndarray]:
+        if name in self.initializers_map:
+            return onnx.numpy_helper.to_array(self.initializers_map[name])
+
+        return None
+
+    def get_conv_params(self, node: onnx.NodeProto) -> ConvParams:
```
fengyuentau, line 173: Could you be part of `ConvParams`?

tools/quantize/block_quantize.py
DaniAffCH: Those parameters were actually not used. I refactored the code to make it more general.

fengyuentau (approved):
LGTM👍 I propose to merge this one after https://github.com/opencv/opencv/pull/25644.

PR #267, opened 2024-07-17
Bugfix: Crash during Evaluation of Yunet on Widerface Dataset

TonyCongqianWang:
With the current opencv-python and numpy packages. Executing the eval script will crash with error:

```
"...\datasets\widerface.py", line 305, in eval
    det = np.append(np.around(det[:, :4], 1), np.around(det[:, -1], 3).reshape(-1, 1), axis=1)
                                                        ~~~^^^^^^^
IndexError: index -1 is out of bounds for axis 1 with size 0
```

The returning `np.empty(shape=(0, 5))` instead of `np.arrayy([])` for empty detections will fix this.

### DIFF
--- a/models/face_detection_yunet/yunet.py
+++ b/models/face_detection_yunet/yunet.py
@@ -52,4 +52,4 @@ class YuNet:
     def infer(self, image):
         # Forward
         faces = self._model.detect(image)
-        return np.array([]) if faces[1] is None else faces[1]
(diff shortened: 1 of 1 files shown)


TonyCongqianWang:
It may be needed to do simlar changes to object_detection_nanodet and object_detection_yolox both classes also use `np.arrayy([])` as default value

fengyuentau (approved):
LGTM 👍 Thank you for the contribution!

PR #268, opened 2024-08-04
[GSoC] Gemm and MatMul block quantization support

DaniAffCH:
This PR enhances the block quantization tool by supporting `Gemm` and `MatMul` layers.

The code has been refactored to make it more general and handle side cases, in particular:

- All the `axis = 1` specific instructions have been modified to support any.
- Handled the case of parameter sharing (when multiple layers share the same set of weights)

The effectiveness and correctness of the tool were tested on the following models:

-  [Vit_b_32](https://dl.opencv.org/models/vit/vit_b_32.onnx) containing both `MatMul` and `Gemm`.
-  [object_tracking_vittrack](https://github.com/opencv/opencv_zoo/blob/main/models/object_tracking_vittrack/object_tracking_vittrack_2023sep.onnx) containing `MatMul`  and parameter sharing across different layers. This model inference has also been tested using OpenCV DNN from the related [demo](https://github.com/opencv/opencv_zoo/blob/main/models/object_tracking_vittrack/demo.cpp), confirming the robustness of the quantized model.

### DIFF
--- a/tools/quantize/block_quantize.py
+++ b/tools/quantize/block_quantize.py
@@ -17,11 +17,7 @@ from onnx import helper
 BITS_TO_NUMPY_TYPE = {8: np.uint8, 16: np.uint16}
 
 
-SUPPORTED_OPS = {
-    "Conv",
-    "Gemm",
-    "MatMul"
-}
+SUPPORTED_OPS = {"Conv", "Gemm", "MatMul"}
 
 ONNX_OPSET = 21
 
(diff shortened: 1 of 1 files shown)


fengyuentau (commented):
Generally looks good to me. Let me test the patch later after.

DaniAffCH:
This gif demonstrates the test performed on vittracker

Blockwise quantization vittracker

fengyuentau (approved):

PR #271, opened 2024-08-29
Change the default value of topK

kaingwade:


fengyuentau:
There is cpp sample that needs to be changed as well.

https://github.com/opencv/opencv_zoo/blob/8bdc41bb5e3db55a99aec2902412f3d930524892/models/person_detection_mediapipe/demo.cpp#L34

fengyuentau (approved):

PR #272, opened 2024-08-29
Added dexined quantized model for edge detection

gursimarsingh:
Added dexined quantized model for edge detection DNN sample.


fengyuentau (commented):
1. License is required for every model in the zoo. Compatible licenses are Apache 2.0, MIT. Please refer to the original license.
2. Add a readme and a link to the sample code in opencv/samples. Also add reference to the original source of the model.
3. Follow the naming pattern. The filename of the model should be `${task}_${model_name}_${timestamp}.onnx`, such as `edge_detection_dexined_2024sep.onnx`.

models/edge_detection_dexined/edge_detection.py
```diff
@@ -0,0 +1,79 @@
...
+        img = np.squeeze(img)
+        img = cv.normalize(img, None, 0, 255, cv.NORM_MINMAX, cv.CV_8U)
+        img = cv.resize(img, (w, h))
+        preds.append(img)
+    fuse = preds[-1]
+    ave = np.array(preds, dtype=np.float32)
+    ave = np.uint8(np.mean(ave, axis=0))
+    return fuse, ave
+
+def loadModel(args):
```
fengyuentau, line 37: Could you write a wrapper class and put it in a separate file? We do this in this way so that we can run benchmark on different models.

Example: https://github.com/opencv/opencv_zoo/blob/main/models/face_detection_yunet/yunet.py

models/edge_detection_dexined/edge_detection.py
```diff
@@ -0,0 +1,79 @@
...
+        inp = cv.dnn.blobFromImage(image, 1.0, (512, 512), (103.5, 116.2, 123.6), swapRB=False, crop=False)
+
+        net.setInput(inp)
+        apply_dexined(net, image)
+
+        cv.imshow("Input", image)
+    cv.destroyAllWindows()
+
+if __name__ == '__main__':
+    main()
```
fengyuentau, line 79: Put an empty line before end of file.

models/edge_detection_dexined/edge_detection.cpp
```diff
@@ -0,0 +1,125 @@
...
+        imshow("Output", result);
+        int key = waitKey(1);
+        if (key == 27 || key == 'q')
+        {
+            break;
+        }
+    }
+    destroyAllWindows();
+    return 0;
+}
```
fengyuentau, line 125: Put an empty line before end of file.

models/edge_detection_dexined/edge_detection.py
gursimarsingh: added a separate file with class for python and added a class in edge_detection.cpp

models/edge_detection_dexined/dexined.py
```diff
@@ -0,0 +1,52 @@
...
+        return fuse, ave
+
+    def infer(self, image):
+        inp = cv.dnn.blobFromImage(image, 1.0, (512, 512), (103.5, 116.2, 123.6), swapRB=False, crop=False)
+        self._model.setInput(inp)
+        
+        # Forward pass through the model
+        out = self._model.forward()
+        result, _ = self.postProcessing(out, image.shape[:2])
+        t, _ = self._model.getPerfProfile()
```
fengyuentau, line 49: No need to get profile data here. The benchmark script measures at the whole `infer` function.

We normally measure the time of the whole pipeline including preprocessing, inference and post processing. You can use `TickMeter` to do so, e.g. https://github.com/opencv/opencv_zoo/blob/c764e507f591f1cce2eb2ee75d0fe5ba7d29388c/models/face_detection_yunet/demo.py#L128-L141

models/edge_detection_dexined/edge_detection.cpp
fengyuentau, line 1: Could you rename the file to `demo.cpp`? Just want to keep aligned with others in this repo.

models/edge_detection_dexined/edge_detection.py
fengyuentau, line 1: Could you rename the file to `demo.py`? Just want to keep aligned with others in this repo.

fengyuentau (approved):
👍

PR #277, opened 2024-11-14
C++ Demo for person_reid_youtureid

redhecker:
## Details
This pull request adds a C++ demo for the person_reid_youtureid requested in #135 
- `README.md` was updated to include basic instructions for compiling and running the C++ demo.
- `CMakeLists.txt` contains the cmake code necessary for building the C++ demo.
- `demo.cpp` contains the source code for the C++ demo.

## Tests
8f21790116b814d59d165c88affdbe0
c3a2d615895f66d46781c129dcb2f1d
dd902b92b68a6ce807964d1ea73e1c7
(test data comes from [https://github.com/ReID-Team/ReID_extra_testdata](https://github.com/ReID-Team/ReID_extra_testdata))

## Notes
one little problem is that cpp17 is needed for directory access

### DIFF
--- a/models/person_reid_youtureid/CMakeLists.txt
+++ b/models/person_reid_youtureid/CMakeLists.txt
@@ -1,14 +1,11 @@
 cmake_minimum_required(VERSION 3.24.0)
 project(opencv_zoo_person_reid_youtureid)
 
-set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 set(OPENCV_VERSION "4.10.0")
 set(OPENCV_INSTALLATION_PATH "" CACHE PATH "Where to look for OpenCV installation")
 
+# Find OpenCV
 find_package(OpenCV ${OPENCV_VERSION} REQUIRED HINTS ${OPENCV_INSTALLATION_PATH})
 
 add_executable(demo demo.cpp)
-target_link_libraries(demo ${OpenCV_LIBS} stdc++fs) 
-
+target_link_libraries(demo ${OpenCV_LIBS})
--- a/models/person_reid_youtureid/CMakeLists.txt
+++ b/models/person_reid_youtureid/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.24.0)
+project(opencv_zoo_person_reid_youtureid)
+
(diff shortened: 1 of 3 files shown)


models/person_reid_youtureid/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.24.0)
+project(opencv_zoo_person_reid_youtureid)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
fengyuentau, line 5: Could you release standard?

OpenCV 4.x stays on C++11, so does the demo.

models/person_reid_youtureid/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+set(OPENCV_VERSION "4.10.0")
+set(OPENCV_INSTALLATION_PATH "" CACHE PATH "Where to look for OpenCV installation")
+
+find_package(OpenCV ${OPENCV_VERSION} REQUIRED HINTS ${OPENCV_INSTALLATION_PATH})
+
+add_executable(demo demo.cpp)
+target_link_libraries(demo ${OpenCV_LIBS} stdc++fs)
```
fengyuentau, line 13: Why "stdc++fs" is needed here?

fengyuentau (changes requested):
Thank you for contributions 👍 Please have a look at the comments below.

models/person_reid_youtureid/CMakeLists.txt
Gongjunzhe12210401: This is because we previously used the filesystem library. We have now made changes and replaced it with OpenCV's built-in method. Now, we only need to use target_link_libraries(demo ${OpenCV_LIBS}).

models/person_reid_youtureid/CMakeLists.txt
Gongjunzhe12210401: We have now updated the code to use C++11 standards.

fengyuentau (approved):
LGTM👍

PR #284, opened 2025-01-08
Added lama quantized model for inpainting DNN sample.

gursimarsingh:
Added lama quantized model for inpainting DNN sample.


fengyuentau (commented):
Thank you for contribution 👍 Just a minor comment below.

Issue #292, opened 2025-04-25
Is license_plate_detection_lpd_yunet_2023mar.onnx CUDA enabled?


cvetelinandreevdreamix:
Hello,

I have open cv installed with CUDA support

  NVIDIA CUDA:                   YES (ver 12.0, CUFFT CUBLAS)
    NVIDIA GPU arch:             86
    NVIDIA PTX archs:

  cuDNN:                         YES (ver 8.9.7)

I'm running both face and plate detection with the models face_detection_yunet_2023mar.onnx and license_plate_detection_lpd_yunet_2023mar.onnx

I'm using the code from the demo files. I'm running the face detection over 10 images with two backend/target configurations and get the following times to get the results:
OpenCV implementation + CPU  = 33 sec 
CUDA + GPU (CUDA) = 7 sec

If I run the plate detection with those configuration I don't get any difference in the processing time + I dont see the process while monitoring the nvidia_smi. It seems that the model is not compatible with CUDA. Is that so?

fengyuentau:
Hello, you can find the benchmark results of the model running on CUDA here: https://github.com/opencv/opencv_zoo/tree/main/benchmark#jetson-nano-b01.

Could you check whether you are using the demo scripts in the right way?

ShiqiYu:
Besides, the plate detection can only work for Chinese plates.

cvetelinandreevdreamix:
Thank you for your answers.

@fengyuentau The way I use the model is almost identical to the demo file. Can't see anything missing there.

@ShiqiYu Plate detection works fine with Bulgarian plates. This model was the best I could find in terms of speed, accuracy and easy of use. Can you recommend something else?

I just tried to run the benchmarks and I get this error

```
python benchmark.py --cfg ./config/face_detection_yunet.yaml
Benchmarking ...
mean       median     min        input size   model
Traceback (most recent call last):
  File "/home/ai/projects/opencv_zoo/benchmark/benchmark.py", line 226, in <module>
    model = model_handler(*model_path, **model_config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ai/projects/opencv_zoo/models/face_detection_yunet/yunet.py", line 22, in __init__
    self._model = cv.FaceDetectorYN.create(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #293, opened 2025-05-07
fix: set backend and target in init

fengyuentau:
Fixes https://github.com/opencv/opencv_zoo/issues/292#issuecomment-2841342795


fengyuentau:
@cvetelinandreevdreamix I made this patch to fix the problem.

PR #304, opened 2026-02-08
DOCS: Improve Model Download Instructions

muhammadawaisofficial:
DOCS: Improve Model Download Instructions


crackwitz:
I don't consider the deletion of 300 files to be valuable.

PR #309, opened 2026-05-05
add INT8 per-channel static PTQ YuNet face detection model 

varun-jaiswal17:
Adds a new INT8 per-channel statically quantized variant of the YuNet
face detection model, produced from the official FP32 model using ONNX
Runtime post-training quantization.

Per-channel quantization assigns one scale/zero-point per output channel
instead of one per entire layer, preserving per-channel weight distribution
and eliminating the accuracy gap seen in per-tensor quantization.

## Conversion script

```python
quantize_static(
    per_channel=True,
    weight_type=QuantType.QInt8,
    activation_type=QuantType.QUInt8,
    quant_format=QuantFormat.QDQ,
    calibrate_method=CalibrationMethod.MinMax,
)


asmorkalov:
Looks like we have to pay for git LFS. I cannot download the model. Do you have it somewhere else? Huggingface?

varun-jaiswal17:
> Looks like we have to pay for git LFS. I cannot download the model. Do you have it somewhere else? Huggingface?

@asmorkalov 
here is the google drive link for the model : https://drive.google.com/file/d/1P5MBGCJADkicXWEFRvDNfaJPkx4ibX7x/view?usp=sharing

